PR #32, opened 2024-12-01
Include val/rdy ifcs and re-organize lib directory

tancheng:
- Include val/rdy ifcs and re-organize lib directory.
 - Include https://github.com/tancheng/PyOCN as sub-module.
   - [ ] Leverage CGRA's lib in PyOCN's ifcs.

### DIFF
--- a/lib/messages.py
+++ b/lib/messages.py
@@ -63,7 +63,8 @@ def mk_predicate( payload_nbits=1, predicate_nbits=1, prefix="CGRAData" ):
 # Generic config message
 #=========================================================================
 
-def mk_ctrl( num_fu_in=2, num_inports=5, num_outports=5, prefix="CGRAConfig" ):
+def mk_ctrl(num_fu_in = 2, num_inports = 5, num_outports = 5,
+            prefix = "CGRAConfig"):
 
   ctrl_nbits    = 6
   CtrlType      = mk_bits( ctrl_nbits )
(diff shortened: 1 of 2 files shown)


tancheng:
@rp15 @yuqisun @yo96 @yyan7223 @MeowMJ plz be aware that I re-organize the lib directory hierarchy and now it contains both `val/rdy` and `en/rdy` ifcs. The change is due to we wanna integrate VectorCGRA with https://github.com/tancheng/PyOCN (@cbatten for awareness), and they use different versions of PyMTL3 (VectorCGRA relies on `en/rdy` in a particular commit: https://github.com/tancheng/pymtl3.1/tree/yo-struct-list-fix supported by @yo96 while PyOCN relies on `val/rdy`, and both `en/rdy` and `val/rdy` don't co-exist in the specific PyMTL3 version), so I include the `val/rdy` in this PR and will change the import path of the included PyOCN to use them (instead of relying on its own PyMTL3 version's [stdlib/stream/ifcs.py](https://github.com/pymtl/pymtl3/blob/master/pymtl3/stdlib/stream/ifcs.py)).
 - Ideally, both should be updated to the latest PyMTL3 and just use `val/rdy`, which could be the long-term goal.
 - For the integration of VectorCGRA and PyOCN, I would keep `en/rdy` within CGRA tiles, while leveraging `val/rdy` to interconnect multiple CGRAs via PyOCN.

yyan7223 (commented):
This is DALAO, much work done in only few days...
But there are too many files changed to review them in a short time, how about we check its functionality in the subsequent use of those functions.

tancheng:
Not really. Most changes are copy-pasted. And PyOCN is also created by @yo96 and me.

True, I am including PyOCN's tests gradually in this PR (WIP). Thanks!

tancheng:
Well, I have not yet completed this PR. I will create a new branch anyways.

yuqisun:
> Well, I have not yet completed this PR. I will create a new branch anyways.

Ah! Sorry I just forgot to switch branch and deleted my change from master then, not sure why merged this.

tancheng:
No worry~!

PR #36, opened 2024-12-03
[feature] Simple controller

tancheng:
A simple controller connected to each CGRA
 - The simple controller interfaces val/rdy with en/rdy ifcs
 - A ring (or any other topology) network would connect to multiple controllers
 - The simple controller only manipulate the data type from/to CGRA to/from ring, while more complicated functionalities can be followed up
   - Dynamic VFS
   - Dynamic rescheduling
   - Dynamic execution time monitoring
   - Invoke/halt specific tiles
     - [ ] The tiles need to be connected to the controller through another NoC instead of via a broadcast bus


controller/ControllerRTL.py
```diff
@@ -0,0 +1,94 @@
...
+    # Can also be @update instead of @update_ff
+    # @update
+    @update_ff
+    def update_controller():
+      s.recv_from_other.rdy <<= s.send_to_master.rdy
+      s.send_to_master.en <<= s.recv_from_other.val & s.send_to_master.rdy
+      s.send_to_master.msg <<= s.pkt2data
+
+      s.recv_from_master.rdy <<= s.send_to_other.rdy
+      s.send_to_other.val <<= s.recv_from_master.en
```
yyan7223, line 84: should it be `s.send_to_other.val <<= s.recv_from_master.en & s.send_to_other.rdy` ?

controller/test/ControllerRTL_test.py
```diff
@@ -0,0 +1,127 @@
...
+    cmp_fn = lambda a, b : a.payload == b.payload
+    s.src_en_rdy = TestSrcRTL(MsgType, src_msgs)
+    s.sink_en_rdy = TestSinkRTL(MsgType, sink_msgs)
+    s.src_val_rdy = TestValRdySrcRTL(PktType, src_pkts)
+    s.sink_val_rdy = TestNetSinkRTL(PktType, sink_pkts, cmp_fn = cmp_fn)
+
+    s.dut = ControllerRTL(PktType, MsgType, AddrType)
+
+    # Connections
+    s.src_en_rdy.send //= s.dut.recv_from_master
```
yyan7223, line 42: Should it be `s.dut.recv_from_master //= s.src_en_rdy.send`,  input signal on the left?

controller/test/ControllerRTL_test.py
```diff
@@ -0,0 +1,127 @@
...
+class TestHarness(Component):
+
+  def construct(s, MsgType, PktType, AddrType, src_msgs, sink_msgs, src_pkts, sink_pkts):
+
+    cmp_fn = lambda a, b : a.payload == b.payload
+    s.src_en_rdy = TestSrcRTL(MsgType, src_msgs)
+    s.sink_en_rdy = TestSinkRTL(MsgType, sink_msgs)
+    s.src_val_rdy = TestValRdySrcRTL(PktType, src_pkts)
+    s.sink_val_rdy = TestNetSinkRTL(PktType, sink_pkts, cmp_fn = cmp_fn)
+
```
yyan7223, line 38: - `TestSrcRTL` simluates how `master` sends data to `controller`.
- `TestSinkRTL` simulates how `master` receives data from `controller`.
- `TestValRdySrcRTL` simluates how `other` sends data to `controller`.
- `TestNetSinkRTL` simulates how `other` receives data from `controller`.

`master` should be a CGRA Tile where `controller` is located, and `other` are multiple `controller` in the other Tiles that are linked with current Tile. Am I correct?

controller/test/ControllerRTL_test.py
```diff
@@ -0,0 +1,127 @@
...
+#-------------------------------------------------------------------------
+# TestHarness
+#-------------------------------------------------------------------------
+
+class TestHarness(Component):
+
+  def construct(s, MsgType, PktType, AddrType, src_msgs, sink_msgs, src_pkts, sink_pkts):
+
+    cmp_fn = lambda a, b : a.payload == b.payload
```
yyan7223, line 33: What is `cmp_fn` used for?

controller/test/ControllerRTL_test.py
```diff
@@ -0,0 +1,127 @@
...
+
+nterminals = 4
+Pkt = mk_ring_multi_cgra_pkt(nterminals, payload_nbits = 32,
+                             predicate_nbits = 1)
+src_pkts = [
+    #   src  dst opq vc payload     predicate
+    Pkt(0,   0,  0,  0, 0xdeadbeef, 0),
+    Pkt(0,   0,  0,  0, 0xbeefdead, 0),
+    Pkt(0,   0,  0,  0, 0xdeedbeed, 0),
+]
```
yyan7223, line 112: Is each `Pkt` corresponding to a Tile in one direction?

controller/ControllerRTL.py
tancheng: `val` means there is data pending to be sent out. `en` is a proactive action that `val` shouldn't depend on it. This line causes hanging, i.e., `val` never be `true`.

I change this by including a `queue` in the latest commit, and change this to `val = queue.count > 0`.

controller/test/ControllerRTL_test.py
tancheng: used to compare with the golden reference, i.e., RingFL.

controller/test/ControllerRTL_test.py
tancheng: From what I understand, when we use `//=`, it means signals on both sides would be driven together (or drive other together). It is `connection` instead of `assignment`.

controller/test/ControllerRTL_test.py
tancheng: Right! Except we need to think about master is one single tile or multiple tiles or the DataMemoryRTL, or all of them. (controller needs to control all of them)

controller/test/ControllerRTL_test.py
tancheng: Corresponding to a `CGRA` (not `tile`) in one (or two) direction. I cannot remember clearly that the ring is uni-direction or bi-direction, plz have a check in the `ringnet/RingNetworkRTL` source code.

yyan7223 (approved):

Issue #26, opened 2024-11-23
Data arrival-dependent execution and cycle-based execution


tancheng:
Make the two execution modes be configurable.

yyan7223:
Oh do you mean the dynamic dataflow like SGMF and static data flow like we've been doing? The dynamic dataflow is awesome but we might replace cache instead of scratchpad memory.

yyan7223:
The prototype that I implement for the MICRO submission uses `Verilog`. I tried `Pymtl3` first but have bugs haven't been fixed. But I believe this issue should have lower priority than #74 cause it relies on the `ld` miss from the data memory.

tancheng:
> The prototype that I implement for the MICRO submission uses `Verilog`. I tried `Pymtl3` first but have bugs haven't been fixed. But I believe this issue should have lower priority than [#74](https://github.com/tancheng/VectorCGRA/issues/74) cause it relies on the `ld` miss from the data memory.

Agree~!

PR #40, opened 2024-12-13
Crossbar based data SRAM

tancheng:
A crossbar is in between the data memory and the tiles.
 - [x] Allows the crossbar to send requests to NoC (via controller).
   - For now, the outstanding request would block all the other remote access requests.
 - [x] Need to check the generated/translated SVerilog.
 ```
        for i in range(num_xbar_in_rd_ports):
          arbitrated_rd_msg = s.read_crossbar.packet_on_input_units[i]
          if (s.read_crossbar.send[arbitrated_rd_msg.dst].msg.src == i) & (arbitrated_rd_msg.dst < num_banks):
            loaded_msg = s.reg_file[trunc(arbitrated_rd_msg.dst, LocalBankIndexType)].rdata[0]
            if i <= s.num_rd_tiles:
```
   - Not sure how synthesizer would handle this, replacing temporary variables with original ones for now.
 - [ ] E2e test
   - Global address space needs to be partitioned across distributed SRAM banks.
   - Larger test, i.e., multi-CGRA with controller and crossbar-based data memories.
     - Checks whether the remote request can get correct response. Can be done in another PR @yuqisun, https://github.com/tancheng/VectorCGRA/issues/44.
 - [x] Different types of requests/messages:
   - Memory access pkt go through the xbar.
   - Load/store requests on NoC.
   - Commands go through NoC.

TODO in next PRs:
 - [ ] Blocking and non-blocking ability (through configuration), https://github.com/tancheng/VectorCGRA/issues/26.


yuqisun (approved):
Cool!

PR #43, opened 2024-12-16
23 simulation traces are not understandable for human being

yuqisun:
Please do NOT merge for now, will add opt value, PR to check build now. Thanks.

### DIFF
--- a/cgra/CGRAKingMeshRTL.py
+++ b/cgra/CGRAKingMeshRTL.py
@@ -153,7 +153,11 @@ class CGRAKingMeshRTL( Component ):
   def line_trace( s ):
     # str = "||".join([ x.element.line_trace() for x in s.tile ])
     # str += " :: [" + s.data_mem.line_trace() + "]"
-    res = "||\n".join([ (("[tile"+str(i)+"]: ") + x.line_trace() + x.ctrl_mem.line_trace())
-                              for (i,x) in enumerate(s.tile) ])
+    # res = "||\n".join([ (("[tile"+str(i)+"]: ") + x.line_trace() + x.ctrl_mem.line_trace())
+    #                           for (i,x) in enumerate(s.tile) ])
+    res = ''
+    for (i, x) in enumerate(s.tile):
+      res += "# [tile"+str(i)+"]: " + x.line_trace() + x.ctrl_mem.line_trace() + '\n'
     res += "\n data_mem:: [" + s.data_mem.line_trace() + "]    \n"
+    res += "------\n\n"
     return res
--- a/cgra/CGRAKingMeshRTL.py
+++ b/cgra/CGRAKingMeshRTL.py
@@ -155,5 +155,5 @@ class CGRAKingMeshRTL( Component ):
     # str += " :: [" + s.data_mem.line_trace() + "]"
     res = "||\n".join([ (("[tile"+str(i)+"]: ") + x.line_trace() + x.ctrl_mem.line_trace())
                               for (i,x) in enumerate(s.tile) ])
-    res += "\n :: [" + s.data_mem.line_trace() + "]    \n"
+    res += "\n data_mem:: [" + s.data_mem.line_trace() + "]    \n"
     return res
--- a/mem/ctrl/CtrlMemRTL.py
+++ b/mem/ctrl/CtrlMemRTL.py
@@ -7,8 +7,7 @@ Control memory for CGRA.
 Author : Cheng Tan
   Date : Dec 21, 2019
 """
-
-
+from py_markdown_table.markdown_table import markdown_table
 from pymtl3 import *
 from pymtl3.stdlib.primitive import RegisterFile
 from ...lib.basic.en_rdy.ifcs import SendIfcRTL, RecvIfcRTL
@@ -73,6 +72,15 @@ class CtrlMemRTL( Component ):
             s.reg_file.raddr[0] <<= s.reg_file.raddr[0] + AddrType( 1 )
 
   def line_trace( s ):
-    out_str  = f'[{", ".join([ str(data.__dict__) for data in s.reg_file.regs ])}]'
-    return f'class: {s.__class__.__name__}, recv_ctrl_msg: {s.recv_ctrl.msg.__dict__} : out: {out_str} : send_ctrl_msg: {str(s.send_ctrl.msg.__dict__)}'
+
+    out_md  = markdown_table([ data.__dict__ for data in s.reg_file.regs ]).set_params(quote=False).get_markdown()
+    # recv_opt_msg = "\n".join([(key + ": " + str(value)) for key, value in recv_opt_msg_dict.items()])
+    recv_ctrl_msg = "\n".join([(key + ": " + str(value)) for key, value in s.recv_ctrl.msg.__dict__.items()])
+    send_ctrl_msg = "\n".join([(key + ": " + str(value)) for key, value in s.send_ctrl.msg.__dict__.items()])
+    return (f'\n## class: {s.__class__.__name__}\n'
+            f'- recv_ctrl_msg:\n'
+            f'{recv_ctrl_msg}\n'
+            f'- out: {out_md}\n'
+            f'- send_ctrl_msg:\n'
+            f'{send_ctrl_msg}\n')
 
--- a/mem/ctrl/CtrlMemRTL.py
+++ b/mem/ctrl/CtrlMemRTL.py
@@ -73,6 +73,6 @@ class CtrlMemRTL( Component ):
             s.reg_file.raddr[0] <<= s.reg_file.raddr[0] + AddrType( 1 )
 
   def line_trace( s ):
-    out_str  = "||".join([ str(data) for data in s.reg_file.regs ])
-    return f'{s.recv_ctrl.msg} : [{out_str}] : {s.send_ctrl.msg}'
+    out_str  = f'[{", ".join([ str(data.__dict__) for data in s.reg_file.regs ])}]'
+    return f'class: {s.__class__.__name__}, recv_ctrl_msg: {s.recv_ctrl.msg.__dict__} : out: {out_str} : send_ctrl_msg: {str(s.send_ctrl.msg.__dict__)}'
 
--- a/mem/data/DataMemRTL.py
+++ b/mem/data/DataMemRTL.py
@@ -99,12 +99,12 @@ class DataMemRTL( Component ):
         s.recv_wdata[i].rdy @= Bits1( 1 )
 
   def line_trace(s):
-    recv_raddr_str = "recv_read_addr: " + "|".join([str(data.msg) for data in s.recv_raddr])
-    recv_waddr_str = "recv_write_addr: " + "|".join([str(data.msg) for data in s.recv_waddr])
-    recv_wdata_str = "recv_write_data: " + "|".join([str(data.msg) for data in s.recv_wdata])
-    content_str = "content: " + "|".join([str(data) for data in s.reg_file.regs])
-    send_rdata_str = "send_read_data: " + "|".join([str(data.msg) for data in s.send_rdata])
-    return f'{recv_raddr_str} || {recv_waddr_str} || {recv_wdata_str} || [{content_str}] || {send_rdata_str}'
+    recv_raddr_str = "recv_read_addr: " + f'[{", ".join([str(data.msg) for data in s.recv_raddr])}]'
+    recv_waddr_str = "recv_write_addr: " + f'[{", ".join([str(data.msg) for data in s.recv_waddr])}]'
+    recv_wdata_str = "recv_write_data: " + f'[{", ".join([str(data.msg.__dict__) for data in s.recv_wdata])}]'
+    content_str = "content: " + f'[{", ".join([str(data.__dict__) for data in s.reg_file.regs])}]'
+    send_rdata_str = "send_read_data: " + f'[{", ".join([str(data.msg) for data in s.send_rdata])}]'
+    return f'class: {s.__class__.__name__}, {recv_raddr_str} || {recv_waddr_str} || {recv_wdata_str} || [{content_str}] || {send_rdata_str}'
     # return f'DataMem: {recv_str} : [{out_str}] : {send_str} initWrites: {s.initWrites}'
     # return s.reg_file.line_trace()
     # return f'<{s.reg_file.wen[0]}>{s.reg_file.waddr[0]}:{s.reg_file.wdata[0]}|{s.reg_file.raddr[0]}:{s.reg_file.rdata[0]}'
--- a/tile/TileRTL.py
+++ b/tile/TileRTL.py
@@ -6,6 +6,7 @@ TileRTL.py
 Author : Cheng Tan
   Date : Dec 11, 2019
 """
+import json
 
 
 from pymtl3 import *
@@ -123,12 +124,13 @@ class TileRTL( Component ):
       s.crossbar.recv_opt.en   @= s.ctrl_mem.send_ctrl.en
       s.ctrl_mem.send_ctrl.rdy @= s.element.recv_opt.rdy & s.crossbar.recv_opt.rdy
 
+
   # Line trace
   def line_trace( s ):
-
-    recv_str    = "|".join([ str(x.msg) for x in s.recv_data ])
-    channel_recv_str = "|".join([ str(x.recv.msg) for x in s.channel ])
-    channel_send_str = "|".join([ str(x.send.msg) for x in s.channel ])
-    out_str  = "|".join([ "("+str(x.msg.payload)+","+str(x.msg.predicate)+")" for x in s.send_data ])
-    return f"{recv_str} => [{s.crossbar.recv_opt.msg}] ({s.element.line_trace()}) => {channel_recv_str} => {channel_send_str} => {out_str}"
+    recv_str    = f'[{", ".join([ str(x.msg.__dict__) for x in s.recv_data ])}]'
+    # recv_json = json.dumps(s.recv_data)
+    channel_recv_str = f'[{", ".join([ str(x.recv.msg.__dict__) for x in s.channel ])}]'
+    channel_send_str = f'[{", ".join([ str(x.send.msg.__dict__) for x in s.channel ])}]'
+    out_str  = f'[{", ".join([ "{send_msg_payload: "+str(x.msg.payload)+", send_msg_predicate: "+str(x.msg.predicate)+"}" for x in s.send_data ])}]'
+    return f"class: {s.__class__.__name__}, recv: {recv_str} => [recv_opt_msg: {str(s.crossbar.recv_opt.msg.__dict__)}] ({s.element.line_trace()}) => channel_recv: {channel_recv_str} => channel_send: {channel_send_str} => out: {out_str}"
 
--- a/tile/TileRTL.py
+++ b/tile/TileRTL.py
@@ -10,6 +10,8 @@ import json
 
 
 from pymtl3 import *
+
+from lib.opt_type import OPT_SYMBOL_DICT
 from ..fu.flexible.FlexibleFuRTL import FlexibleFuRTL
 from ..fu.single.AdderRTL import AdderRTL
 from ..fu.single.BranchRTL import BranchRTL
@@ -24,7 +26,10 @@ from ..noc.CrossbarRTL import CrossbarRTL
 from ..noc.ChannelRTL import ChannelRTL
 from ..rf.RegisterRTL import RegisterRTL
 # from ..noc.BypassChannelRTL      import BypassChannelRTL
+from py_markdown_table.markdown_table import markdown_table
 
+tile_port_direction_dict = {0: "NORTH", 1: "SOUTH", 2: "WEST", 3: "EAST", 4: "NORTHWEST", 5: "NORTHEAST",
+                              6: "SOUTHEAST", 7: "SOUTHWEST"}
 
 class TileRTL( Component ):
 
@@ -127,10 +132,40 @@ class TileRTL( Component ):
 
   # Line trace
   def line_trace( s ):
-    recv_str    = f'[{", ".join([ str(x.msg.__dict__) for x in s.recv_data ])}]'
-    # recv_json = json.dumps(s.recv_data)
-    channel_recv_str = f'[{", ".join([ str(x.recv.msg.__dict__) for x in s.channel ])}]'
-    channel_send_str = f'[{", ".join([ str(x.send.msg.__dict__) for x in s.channel ])}]'
-    out_str  = f'[{", ".join([ "{send_msg_payload: "+str(x.msg.payload)+", send_msg_predicate: "+str(x.msg.predicate)+"}" for x in s.send_data ])}]'
-    return f"class: {s.__class__.__name__}, recv: {recv_str} => [recv_opt_msg: {str(s.crossbar.recv_opt.msg.__dict__)}] ({s.element.line_trace()}) => channel_recv: {channel_recv_str} => channel_send: {channel_send_str} => out: {out_str}"
+    # recv_str    = f'[{", ".join([ str(x.msg.__dict__) for x in s.recv_data ])}]'
+    recv_data = [ x.msg.__dict__ for x in s.recv_data ]
+    recv_list = []
+    for idx, data in enumerate(recv_data):
+      port_direction = tile_port_direction_dict[idx]
+      dict_with_direction = {"port_direction": port_direction}
+      dict_with_direction.update(data)
+      recv_list.append(dict_with_direction)
+    recv_md = markdown_table(recv_list).set_params(quote=False).get_markdown()
+    recv_opt_msg_dict = s.crossbar.recv_opt.msg.__dict__
+    # todo
+    # E   AttributeError: 'str' object has no attribute 'clone'
+    # print(f"s.crossbar.recv_opt.msg: {s.crossbar.recv_opt.msg}")
+    # recv_opt_msg_dict['ctrl'] = OPT_SYMBOL_DICT[s.crossbar.recv_opt.msg.ctrl]
+    recv_opt_msg = "\n".join([(key + ": " + str(value)) for key, value in recv_opt_msg_dict.items()])
+
+    channel_recv_md = markdown_table([ x.recv.msg.__dict__ for x in s.channel ]).set_params(quote=False).get_markdown()
+    channel_send_md = markdown_table([ x.send.msg.__dict__ for x in s.channel ]).set_params(quote=False).get_markdown()
+    out_md  = markdown_table([ dict(send_msg_payload=x.msg.payload, send_msg_predicate=x.msg.predicate) for x in s.send_data ]).set_params(quote=False).get_markdown()
+    return (f"\n## class[{s.__class__.__name__}]:\n"
+            f"- recv:"
+            f"{recv_md}\n"
+            f"===>\n"
+            f"- recv_opt_msg:\n"
+            f"{recv_opt_msg}\n"
+            f"- element:\n"
+            f"{s.element.line_trace()}\n"
+            f"===>\n"
+            f"- channel_recv:\n"
+            f"{channel_recv_md}\n"
+            f"===>\n"
+            f"- channel_send:"
+            f"{channel_send_md}\n"
+            f"===>\n"
+            f"- out:"
+            f"{out_md}\n")
 
--- a/fu/flexible/FlexibleFuRTL.py
+++ b/fu/flexible/FlexibleFuRTL.py
@@ -8,6 +8,7 @@ Author : Cheng Tan
   Date : Dec 24, 2019
 
 """
+import json
 
 from pymtl3 import *
 from ...fu.single.MemUnitRTL import MemUnitRTL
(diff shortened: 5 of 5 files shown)


tancheng:
Hi @yuqisun, thanks for the PR,
```
+--------------+----------------+---------+------+-----+
|port_direction|     payload    |predicate|bypass|delay|
+--------------+----------------+---------+------+-----+
|     NORTH    |0000000000000000|    0    |   0  |  0  |
+--------------+----------------+---------+------+-----+
```
- Can we distinguish the `port_direction` into `inport_direction` and `outport_direction`?
```
outport: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
predicate_in: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
```
 - Can we have the meaningful content inside those array? Say, `outport: [N, S, -, -, ..., -]`?

```
===>
- channel_recv:

+----------------+---------+------+-----+
|     payload    |predicate|bypass|delay|
+----------------+---------+------+-----+
|0000000000000000|    0    |   0  |  0  |
+----------------+---------+------+-----+
|0000000000000000|    0    |   0  |  0  |
+----------------+---------+------+-----+
```
 - What is above `channel_recv`, is this necessary to be shown? If yes, can we provide its name?

yuqisun:
Hi,
`channel_recv` is from:
`[ x.recv.msg.__dict__ for x in s.channel ]` <=== `s.channel  = [ ChannelRTL( DataType ) for _ in range( num_xbar_outports ) ] `<=== `num_xbar_outports = num_fu_inports + num_connect_outports`.

So channel should be the line of crossbar, channel_recv should be the inports which in green in below picture, am I correct?

tancheng:
> Hi,
> 
> `channel_recv` is from: `[ x.recv.msg.__dict__ for x in s.channel ]` <=== `s.channel = [ ChannelRTL( DataType ) for _ in range( num_xbar_outports ) ] `<=== `num_xbar_outports = num_fu_inports + num_connect_outports`.
> 
> So channel should be the lines of crossbar, channel_recv should be the inports which in green in below picture, am I correct?
> 
> image
> 
> And there's also `channel_send`: `[ x.send.msg.__dict__ for x in s.channel ]` which should be the outports in blue in picture?

Nope. `Channel` actually includes registers. The `channel.recv` is the blue one, the `channel.send` is the one after the yellow register. Imagine the `channel` as a `pipe`, it doesn't contain the xbar, it connects to the xbar.

yuqisun:
Hi Cheng,

Updated format in attached log. Mainly 4 parts:
1. Tile (in and out)
2. FU (opt, in and out)
3. ctrl mem (in, out and regs)
4. data mem (regs and read/write addr/data)

[VectorCGRAKingMeshRTL_test_log3.txt](https://github.com/user-attachments/files/18221792/VectorCGRAKingMeshRTL_test_log3.txt)

------

I draft a structure graph [here](https://excalidraw.com/#json=3tWjRiZQv9k6IUWTzex0O,zHHnV-NweJjawVvFYQRbMw) based on my understanding, could you help review once get time, and have some questions about the ctrl mem and data mem:
1. ctrl mem
  a) CtrlMem is ConfigMem in paper? Is it used to control the tile, FU?
  b) There're recv_ctrl_msg and send_ctrl_msg, is the send_ctrl_msg the values send to fu in and next tile in? and recv_ctrl_msg the values from fu out and other tile out?
  c) regs are used to run one by once clk by clk?

2. data mem
  There're raddr, rdata, waddr, wdata, are they in different clk?

Appreciate it if can have a talk sometime.

Thanks,

tancheng:
Thanks @yuqisun,

For the `reg` format:
 - Can the `|(start)|    0    |[0, 0, 0, 0]|['N(-)', 'S(-)', 'W(-)', 'E(-)', 'NW(-)', 'NE(-)', 'SE(-)', 'SW(-)', 'fu_reg_1(-)', 'fu_reg_2(-)', 'fu_reg_3(-)', 'fu_reg_4(-)']|['N(0)', 'S(0)', 'W(0)', 'E(0)', 'NW(0)', 'NE(0)', 'SE(0)', 'SW(0)', 'fu_out_1(0)', 'fu_out_2(0)']|
+-------+---------+------------+--------------------------------------------------------------------------------------------------------------------------------+--------------------------------------------------------------------------------------------------+` be also in a more dense format? i.e.,
```
+-------+---------+---------------+-----------------------------------------------+...
|  ctrl |predicate|    fu_in      |                outport                        |                                           predicate_in                                           |
+-------+---------+---------------+-----------------------------------------------+...
|       |         | 0 | 1 | 2 | 3 | N | S | W | E | NW | NE | SE | ... | fu_reg_1 |                             |                                           predicate_in                                           |...
+-------+---------+---------------+-----------------------------------------------+...
|(start)|    0    | - | 1 | - | - | - | N | S | - | -  | -  | N  | ... |    SW    |                             |...
+-------+---------+---------------+-----------------------------------------------+...
```
 - Similar requirement can be applied on the `send_ctrl_msg`.

To answer ur questions:
 - Ctrl mem:
   -  Yes, exactly. It is also used to control the xbar.
   - If you are talking about: https://github.com/tancheng/VectorCGRA/blob/d1dd6b300e6848cbcde7cff6c6e5f7131428b01c/mem/ctrl/CtrlMemRTL.py#L38
     - It is a bit confusing. `recv_ctrl` is for preloading the control signals into the `ctrl_memory` **before** the kernel is being launched/accelerated. `send_ctrl` is for guiding what should be done on the **local** tile's FU/xbar **once** the kernel is launched.
     - For example, each tile can `recv` 4 control signals and put them into their `ctrl_memory` using 4 cycles. Then, the kernel can start by iterating the 4 control signals repeatedly, each cycle 1 control signal is `sent` to the FU/xbar. The ctrl signals resident inside `tile0` would not affect other tiles' (e.g., `tile1`) behavior. We can sync on this part if it is still not clear to you.
   - If you are referring `regs` to: https://github.com/tancheng/VectorCGRA/blob/d1dd6b300e6848cbcde7cff6c6e5f7131428b01c/mem/ctrl/CtrlMemRTL.py#L41
     - It is actually the `ctrl_memory`. We use this `register_file` to hold the control signals. When we write control signals into it before launching kernel, we need provide address and control signal content (via `recv_ctrl`).
     - For example, a kernel with only two control signals, `cycle0`, we write `add` into `address0`; `cycle1`, we write `mul` into `address1`. When we read control signals from it (via `send_ctrl`) during accelerating the kernel: `cycle0`, we read `add` from `address0`; `cycle1`, we read `mul` from `address1`; `cycle2`, we read `add` from `address0`; `cycle3`, we read `mul` from `address1`; repeating above...
 - Data mem:
   - Read and write can be separate and independent to each other, and they can happen at the same cycle.

yuqisun:
New format attached. Sub header added, please help review.
[VectorCGRAKingMeshRTL_test_log_20241223.txt](https://github.com/user-attachments/files/18227672/VectorCGRAKingMeshRTL_test_log_20241223.txt)

Screenshot from 2024-12-23 16-24-22
Screenshot from 2024-12-23 16-24-53

Thanks Cheng. Matched with what you taught before, each ctrl_memory saves all opts for each tile, like below: ctrl_memory of tile0 saves [LD, PTR] opts, tile1 saves [EXT, LD]
Screenshot from 2024-12-23 17-07-02

tancheng:
@yo96 How do you think the dumped trace? Is it still okay to be called `line_trace()`?

yo96:
The format looks great! Just a quick thought, it looks more like a "trace" than a "line trace," since "line trace" refers to a single-line trace. That said, it's perfectly fine to include this trace in the `line_trace` method if it helps with debugging.

It might still be helpful to have a true line trace as well, as it makes it easier to visualize what's happening across different clock cycles. One idea could be to add a parameter to the `line_trace` method and use the `set_param` method to select trace verbosity during the simulation.

For example:
```
...
  def line_trace(verbosity=0):
    if verbosity = 0:
      return f'{self.in_}(){self.out}'
    if verbosity = 1:
      return self.verbose_trace()

dut = Top()
dut.set_param('top.x.y.z.line_trace', verbosity=1)
...
```
Might not exactly work as-is, but that's the idea.

yuqisun:
Thanks Ou. Working on it.

Hi both,

Got a question about implementation of input verbosity, for the input, we'd better have a command argument to input verbosity level right (such as --test-verilog ...), but default command addoption is in pymtl right? So are we good with manually change verbosity level in *_test.py code when a verbose trace is required? Or it's worth to create another branch for pymtl?

Thanks,

tancheng:
Hi @yuqisun, thanks!

GEMINI said PyMTL doesn't have built-in verbosity command/flag (@yo96 correct me if I am wrong):
Screenshot 2024-12-24 at 6 40 21 PM

It looks good to me if users set the required `verbosity` [here](https://github.com/tancheng/VectorCGRA/blob/49a40187f90a8d67b0a410c0d3bc5944422e22e6/systolic/translate/CgraMemRightAndBottomRTL_matmul_2x2_test.py#L83) or use the `set_param()` as @yo96 mentioned.

@yuqisun You mentioned you would like to change the `line_trace()` to `trace()`, right?

yuqisun:
Yes, then should be good.

For func name: `line_trace` is the fixed name for [pymtl](https://github.com/tancheng/pymtl3.1/blob/6db7736ec52e19cbcfc4adeb7868cae890316591/pymtl3/passes/sim/PrepareSimPass.py#L81) to check attribute, so may not change, I added `verbose_trace` and branch in `line_trace`

```
if self.print_line_trace and hasattr( top, 'line_trace' ):
      final_schedule.append( top.print_line_trace )
```

tancheng:
Plz also provide brief description in this PR, about how to use the original line_trace() and ur improved/modified version in `xxx_test.py`.

yuqisun:
### How to use `line_trace` and `verbose_trace` in `*_test.py`
There're 2 kinds of formats of trace log:

1. `line_trace`: refers to a single-line trace
2. `verbose_trace`: prints more details, such as ctrl_mem ...

You can set `verbosity` [here](https://github.com/tancheng/VectorCGRA/blob/49a40187f90a8d67b0a410c0d3bc5944422e22e6/systolic/translate/CgraMemRightAndBottomRTL_matmul_2x2_test.py#L83), default is zero which uses line_trace.
Set `verbosity=1` will activate verbose logs if there's `verbose_trace` function in class.

For some classes, i.e. `CtrlMemRTL`, there're 3 levels, verbosity 0, 1, and 2, the greater the more details.

**Note:** Please don't use `--test-verilog` when test as it will use `line_trace` in python file which is generated by PyMTL instead of defined in `VectorCGRA`.

.github/workflows/python-package.yml
```diff
@@ -52,20 +52,34 @@ jobs:
...
+        # 3x2 CGRA performs 2x2 matmul translation.
         pytest ../cgra/translate/CGRAMemBottomRTL_matmul_2x2_test.py -xvs --tb=short --test-verilog --dump-vtb --dump-vcd
+        # 3x3 CGRA performs 2x2 matmul simulation/translation.
         pytest ../cgra/translate/CGRAMemRightAndBottomRTL_matmul_2x2_test.py -xvs --tb=short
         pytest ../cgra/translate/CGRAMemRightAndBottomRTL_matmul_2x2_test.py -xvs --tb=short --test-verilog --dump-vtb --dump-vcd
+        # Ring network simulation.
         pytest ../noc/PyOCN/pymtl3_net/ringnet/test/RingNetworkRTL_test.py --tb=short -sv
-        pytest --tb=short -sv ../cgra/translate/CGRAWithControllerRTL_test.py --test-verilog --dump-vtb --dump-vcd
+        # CGRA with separate crossbars (for tiles and FUs), crossbar-based data
+        # memory (for multi-bank), and controller.
```
tancheng, line 79: Did u already sync with TOT/master? Some tests already been deleted by my recent PR.

cgra/CGRAKingMeshRTL.py
```diff
@@ -149,11 +149,23 @@ def construct( s, DataType, PredicateType, CtrlType, width, height,
...
-    # str = "||".join([ x.element.line_trace() for x in s.tile ])
-    # str += " :: [" + s.data_mem.line_trace() + "]"
-    res = "||\n".join([ (("[tile"+str(i)+"]: ") + x.line_trace() + x.ctrl_mem.line_trace())
-                              for (i,x) in enumerate(s.tile) ])
-    res += "\n :: [" + s.data_mem.line_trace() + "]    \n"
+  # verbose trace if verbosity > 0
+  def verbose_trace(s, verbosity=1):
+    res = ''
+    for (i, x) in enumerate(s.tile):
+      res += "# [tile" + str(i) + "]: " + x.line_trace(verbosity=verbosity) + x.ctrl_mem.line_trace(verbosity=verbosity) + '\n'
```
tancheng, line 156: I didn't get this part. It looks if `verbose_trace(1)`, this line would call `line_trace(1)`, then `line_trace(1)` would call `verbose_trace(1)`?

tancheng (commented):
- Plz sync to TOT
- Plz enable the `verbosity` for https://github.com/tancheng/VectorCGRA/blob/master/scale_out/RingMultiCgraRingCtrlMemRTL.py & https://github.com/tancheng/VectorCGRA/blob/master/systolic/CgraMemRightAndBottomRTL.py

Thanks!

cgra/CGRAKingMeshRTL.py
yuqisun: Yes, not all classes have `verbose_trace` now, if there is, `line_trace `will call `verbose_trace`, or will run `line_trace` itself.

.github/workflows/python-package.yml
yuqisun: let me sync

cgra/CGRAKingMeshRTL.py
tancheng: Can the `verbose_trace()` just have:
```
res += "# [tile" + str(i) + "]: " + x.verbose_trace(verbosity=verbosity) + x.ctrl_mem.verbose_trace(verbosity=verbosity) + '\n'
```
instead of 
```
res += "# [tile" + str(i) + "]: " + x.line_trace(verbosity=verbosity) + x.ctrl_mem.line_trace(verbosity=verbosity) + '\n'
```
which is more intuitive?
For the modules/classes that haven't designed `verbose_trace()`, you can have:
```
def verbose_trace(s, verbosity = 1):
  s.line_trace(verbosity)
```
WDYT?

cgra/CGRAKingMeshRTL.py
```diff
@@ -149,11 +149,23 @@ def construct( s, DataType, PredicateType, CtrlType, width, height,
...
+    res += f"\ndata_mem: {s.data_mem.line_trace(verbosity=verbosity)}"
+    res += "------\n\n"
     return res
+
+  # Line trace
+  def line_trace(s, verbosity=0):
+    if verbosity==0:
+      # str = "||".join([ x.element.line_trace() for x in s.tile ])
```
tancheng, line 164: Space before and after all operators, e.g., `verbosity=0` -> ` verbosity = 0 `, `verbosity==0` -> `verbosity == 0`. Plz all this to all the changes.

cgra/CGRAKingMeshRTL.py
yuqisun: Or it's better to add verbose trace to every class?

cgra/CGRAKingMeshRTL.py
yuqisun: Got it.

mem/ctrl/CtrlMemRTL.py
```diff
@@ -72,7 +73,224 @@ def update_raddr():
           else:
             s.reg_file.raddr[0] <<= s.reg_file.raddr[0] + AddrType( 1 )
 
-  def line_trace( s ):
-    out_str  = "||".join([ str(data) for data in s.reg_file.regs ])
-    return f'{s.recv_ctrl.msg} : [{out_str}] : {s.send_ctrl.msg}'
+  # verbose trace if verbosity > 0
```
tancheng, line 76: Is that possible to extract some logic from the huge verbose_trace() func into the lib folder, which can then be used by other modules? It looks super huge now.

tile/TileRTL.py
```diff
@@ -23,7 +24,7 @@
 from ..noc.ChannelRTL import ChannelRTL
 from ..rf.RegisterRTL import RegisterRTL
 # from ..noc.BypassChannelRTL      import BypassChannelRTL
-
+from py_markdown_table.markdown_table import markdown_table
```
tancheng, line 27: Move this after `from pymtl3 import *`, we are sorting these in alphabet ordering.

tile/TileRTL.py
```diff
@@ -123,12 +124,50 @@ def update_opt():
       s.crossbar.recv_opt.en   @= s.ctrl_mem.send_ctrl.en
       s.ctrl_mem.send_ctrl.rdy @= s.element.recv_opt.rdy & s.crossbar.recv_opt.rdy
 
-  # Line trace
-  def line_trace( s ):
+  # verbose trace if verbosity > 0
```
tancheng, line 127: Similar to the above `extract common part` review...

tile/TileRTL_constant.py
```diff
@@ -0,0 +1,21 @@
+tile_port_direction_dict = {
```
tancheng, line 1: Remove this file and add the two dicts in https://github.com/tancheng/VectorCGRA/blob/master/lib/util/common.py? And use `PORT_xxx` as key?

tile/TileSeparateCrossbarRTL.py
```diff
@@ -13,10 +13,11 @@
 Author : Cheng Tan
   Date : Nov 26, 2024
 """
-
-
+from py_markdown_table.markdown_table import markdown_table
```
tancheng, line 16: move after `from pymtl3 ...`

tile/TileSeparateCrossbarRTL.py
```diff
@@ -194,15 +195,115 @@ def update_opt():
                                   s.routing_crossbar.recv_opt.rdy & \
                                   s.fu_crossbar.recv_opt.rdy
 
-
+  # verbose trace if verbosity > 0
```
tancheng, line 198: Extract common part?

cgra/CGRAKingMeshRTL.py
```diff
@@ -149,11 +149,26 @@ def construct( s, DataType, PredicateType, CtrlType, width, height,
...
     # str = "||".join([ x.element.line_trace() for x in s.tile ])
     # str += " :: [" + s.data_mem.line_trace() + "]"
-    res = "||\n".join([ (("[tile"+str(i)+"]: ") + x.line_trace() + x.ctrl_mem.line_trace())
-                              for (i,x) in enumerate(s.tile) ])
+    res = "||\n".join([(("[tile" + str(i) + "]: ") + x.line_trace() + x.ctrl_mem.line_trace())
+                       for (i, x) in enumerate(s.tile)])
     res += "\n :: [" + s.data_mem.line_trace() + "]    \n"
     return res
+
+  def trace(self, verbosity = 0):
```
yuqisun, line 170: How about now, add a trace entrance, but in *_test.py, function name has to be line_trace(). @tancheng

cgra/CGRAKingMeshRTL.py
yuqisun: How about now, add a trace entrance, but in *_test.py, function name has to be line_trace().

mem/ctrl/CtrlMemRTL.py
yuqisun: Yes, think the same thing when I wrote them.

cgra/CGRAKingMeshRTL.py
tancheng: [This](https://github.com/tancheng/VectorCGRA/blob/49a40187f90a8d67b0a410c0d3bc5944422e22e6/cgra/test/CGRAKingMeshRTL_test.py#L61) has to be line_trace(), while [this](https://github.com/tancheng/VectorCGRA/blob/49a40187f90a8d67b0a410c0d3bc5944422e22e6/cgra/test/CGRAKingMeshRTL_test.py#L62) could just be trace(), right?

yuqisun: Oops, ignore above, build error when PyMTL trys to generate .py, *_test.py has to call method line_trace() ... Let me revert commit tmr and change as you mentioned.

cgra/CGRAKingMeshRTL.py
yuqisun: done

cgra/CGRAKingMeshRTL.py
yuqisun: done

tile/TileRTL.py
yuqisun: Done, py_ should be before pymtl3 when optimize.

tile/TileRTL_constant.py
yuqisun: done

tile/TileSeparateCrossbarRTL.py
yuqisun: done

tile/TileRTL.py
yuqisun: done

tile/TileSeparateCrossbarRTL.py
yuqisun: done

mem/ctrl/CtrlMemRTL.py
yuqisun: done

yuqisun:
> * Plz sync to TOT
> 
>     * Plz enable the `verbosity` for https://github.com/tancheng/VectorCGRA/blob/master/scale_out/RingMultiCgraRingCtrlMemRTL.py & https://github.com/tancheng/VectorCGRA/blob/master/systolic/CgraMemRightAndBottomRTL.py
> 
> 
> Thanks!

CgraMemRightAndBottomRTL good.
RingMultiCgraRingCtrlMemRTL pending on CtrlMemDynamicRTL and DataMemWithCrossbarRTL, will add tmr.

lib/util/common.py
```diff
@@ -21,3 +21,25 @@
 LINK_NO_MEM   = 0
 LINK_FROM_MEM = 1
 LINK_TO_MEM   = 2
+
+TILE_PORT_DIRECTION_DICT = {
```
tancheng, line 25: `TILE_PORT_DIRECTION_DICT` -> `TILE_PORT_DIRECTION_DICT_DESC`?

mem/ctrl/CtrlMemRTL.py
```diff
@@ -72,7 +72,121 @@ def update_raddr():
...
+  def verbose_trace( s, verbosity = 1 ):
+    num_fu_in = len(s.reg_file.regs[0].fu_in)
+    # num_inports = len(s.reg_file.regs[0].predicate_in)
+    num_outports = len(s.reg_file.regs[0].outport if hasattr(s.reg_file.regs[0], 'outport') else [])
+    num_direction_ports = num_outports - num_fu_in
+
+    # recv_ctrl
+    recv_ctrl_msg_dict = dict(s.recv_ctrl.msg.__dict__)
+    recv_ctrl_sub_header = s.verbose_trace_data_processor(recv_ctrl_msg_dict, num_direction_ports)
+    recv_ctrl_msg_list = [ recv_ctrl_sub_header, recv_ctrl_msg_dict ]
```
tancheng, line 157: `recv_ctrl_msg_list = [ recv_ctrl_sub_header, recv_ctrl_msg_dict ]` -> `recv_ctrl_msg_list = [recv_ctrl_sub_header, recv_ctrl_msg_dict]`

tile/TileRTL.py
```diff
@@ -123,12 +125,47 @@ def update_opt():
...
+              f"- Tile recv:"
+              f"{recv_md}\n\n"
+              f"- FU element:\n"
+              f"{s.element.verbose_trace(verbosity=verbosity)}\n"
+              f"===>\n"
+              f"- Tile out:"
+              f"{send_md}\n")
+
   # Line trace
   def line_trace( s ):
```
tancheng, line 165: Let's put `line_trace()` before `verbose()`.

yuqisun: done

lib/util/common.py
yuqisun: done

mem/ctrl/CtrlMemRTL.py
yuqisun: done

mem/data/DataMemRTL.py
```diff
@@ -98,7 +97,7 @@ def update_signal():
         s.recv_waddr[i].rdy @= Bits1( 1 )
         s.recv_wdata[i].rdy @= Bits1( 1 )
 
-  def line_trace(s):
+  def line_trace( s ):
```
tancheng, line 100: Remove space.

systolic/CgraMemRightAndBottomRTL.py
```diff
@@ -124,11 +124,21 @@ def construct(s, DataType, PredicateType, CtrlType, width, height,
...
     # str = "||".join([ x.element.line_trace() for x in s.tile ])
     # str += " :: [" + s.data_mem.line_trace() + "]"
     res = "||\n".join([(("[tile" + str(i) + "]: ") + x.line_trace() + x.ctrl_mem.line_trace())
                        for (i,x) in enumerate(s.tile)])
     res += "\n :: SouthMem [" + s.data_mem_south.line_trace() + "]    \n"
     res += "\n :: EastMem [" + s.data_mem_east.line_trace() + "]    \n"
     return res
+
+
+  def verbose_trace( s, verbosity = 1 ):
```
tancheng, line 137: Remove the unnecessary space, (keep the one before and after `=`).

systolic/CgraMemRightAndBottomRTL.py
```diff
@@ -124,11 +124,21 @@ def construct(s, DataType, PredicateType, CtrlType, width, height,
         s.tile[i].to_mem_wdata.rdy //= 0
 
   # Line trace
-  def line_trace(s):
+  def line_trace( s ):
```
tancheng, line 127: Remove unnecessary spaces.

scale_out/test/RingMultiCgraRingCtrlMemRTL_test.py
```diff
@@ -81,6 +81,7 @@ def done(s):
     # return True
 
   def line_trace(s):
+    # return s.dut.verbose_trace(verbosity = 2)
```
tancheng, line 84: If this is the way to invoke the verbose_trace(), let's enable it by default for this test?

tile/test/TileSeparateCrossbarRTL_test.py
```diff
@@ -81,6 +81,7 @@ def done(s):
     return True
 
   def line_trace(s):
+    # return s.dut.verbose_trace(verbosity = 2)
```
tancheng, line 84: Enable it by default for this test?

systolic/translate/CgraMemRightAndBottomRTL_matmul_2x2_test.py
```diff
@@ -80,6 +80,7 @@ def check_parity(s):
     return True
 
   def line_trace(s):
+    # return s.dut.verbose_trace(verbosity = 2)
```
tancheng, line 83: Enable by default?

tancheng (approved):

tancheng:
Hi @yuqisun, I am currently working on a huge cleanup. Plz let me merge first... So then you can sync and merge. Sorry about the inconvenience.

You can move to https://github.com/tancheng/VectorCGRA/issues/44, thanks~!

mem/data/DataMemRTL.py
yuqisun: done

systolic/CgraMemRightAndBottomRTL.py
yuqisun: done

systolic/CgraMemRightAndBottomRTL.py
yuqisun: done

scale_out/test/RingMultiCgraRingCtrlMemRTL_test.py
yuqisun: verbose_trace works for `-xvs --tb=short` but not `--test-verilog`, it looks up `line_trace()`, can test this with `-xvs --tb=short` for RingMultiCgraRingCtrlMemRTL_test.py, current is [`pytest --tb=short -sv ../scale_out/translate/RingMultiCgraRingCtrlMemRTL_test.py --test-verilog --dump-vtb --dump-vcd`](https://github.com/tancheng/VectorCGRA/blob/49a40187f90a8d67b0a410c0d3bc5944422e22e6/.github/workflows/python-package.yml#L82)?

tile/test/TileSeparateCrossbarRTL_test.py
yuqisun: done

systolic/translate/CgraMemRightAndBottomRTL_matmul_2x2_test.py
yuqisun: There're both [`-xvs --tb=short`](https://github.com/tancheng/VectorCGRA/blob/49a40187f90a8d67b0a410c0d3bc5944422e22e6/.github/workflows/python-package.yml#L72) and [`--tb=short --test-verilog --dump-vtb --dump-vcd`](https://github.com/tancheng/VectorCGRA/blob/49a40187f90a8d67b0a410c0d3bc5944422e22e6/.github/workflows/python-package.yml#L73C87-L73C134) in testcases, if enable verbose_trace by default, can I remove and ignore `--tb=short --test-verilog --dump-vtb --dump-vcd` case in action workflow?

scale_out/test/RingMultiCgraRingCtrlMemRTL_test.py
tancheng: - Then the `xxx_test.py` within `translate/` folder can have `line_trace()`, while the one in `test/` can have vervose_trace() enabled by default.

 - Or I guess actually you can control it based on the flag?

```
  def line_trace(s, verbosity=0):
    if verbosity = 0:
      return s.dut.line_trace(verbosity)
    if verbosity = 1:
      return s.dut.verbose_trace(verbosity)

......

dut = Top()

if flags not contains --test-verilog:
  dut.set_param('top.x.y.z.line_trace', verbosity=1) # Or use a global variable?
```

yuqisun: Oops, you both are right, controlling by `set_param` is better, I should use it from the beginning.

Added a line in *_test.py `th.set_param('top.dut.line_trace', verbosity = 1)`, will enable verbose_trace by default and when there's `--test-verilog`, will ignore verbose itself.

systolic/translate/CgraMemRightAndBottomRTL_matmul_2x2_test.py
yuqisun: Make it by `set_param`.

PR #61, opened 2025-01-08
update data preloading

yyan7223:
example

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -52,6 +52,7 @@ jobs:
         pip install -U git+https://github.com/tancheng/pymtl3.1@yo-struct-list-fix
         pip install hypothesis
         pip install pytest
+        pip install py-markdown-table
         pip list
         
     - name: Test and translate with pytest
@@ -77,4 +78,5 @@ jobs:
         pytest ../scale_out/test/RingMultiCgraRTL_test.py -xvs --test-verilog --dump-vtb --dump-vcd
         # Multi-cgra with mesh topology.
         pytest ../scale_out/test/MeshMultiCgraRTL_test.py -xvs --test-verilog --dump-vtb --dump-vcd
-
+        # Const Queue
+        pytest ../mem/const/test/ConstQueueDynamicRTL_test.py -xvs
--- a/mem/const/ConstQueueDynamicRTL.py
+++ b/mem/const/ConstQueueDynamicRTL.py
@@ -1,0 +1,113 @@
+"""
+==========================================================================
+ConstQueueDynamicRTL.py
+==========================================================================
+Constant Queue with regs used for simulation.
+If queue is full, will stop receiving new data.
+
+Author : Yuqi Sun
+  Date : Jan 11, 2025
+"""
+from py_markdown_table.markdown_table import markdown_table
+from pymtl3.stdlib.primitive import RegisterFile
+
+from ...lib.basic.val_rdy.ifcs import ValRdyRecvIfcRTL as RecvIfcRTL
+from ...lib.basic.val_rdy.ifcs import ValRdySendIfcRTL as SendIfcRTL
+from ...lib.opt_type import *
+
+
+class ConstQueueDynamicRTL(Component):
+  def construct(s, DataType, const_mem_size):
+    # Constant
+    # addr type: number of bits to represent the address
+    # 2^addr_size = const_mem_size
+    # i.e. const_mem_size = 8
+    # ConstMemAddrType is 3 bits
+    # ConstMemAddrTyp(1) = 001
+    # ConstMemAddrType(2) = 010
+    AddrType = mk_bits(max(1, clog2(const_mem_size)))
+
+    # Makes write cursor type 1 bit more than mem addr type as need compare with const_mem_size,
+    # otherwise, number will be back to 000 when 111 + 1 (given const_mem_size = 8)
+    WrCurType = mk_bits(clog2(const_mem_size + 1))
+
+    # write cursor and read cursor
+    s.wr_cur = Wire(WrCurType)
+    s.rd_cur = Wire(AddrType)
+
+    # Interface
+    s.send_const = SendIfcRTL(DataType)
+    s.recv_const = RecvIfcRTL(DataType)
+
+    # Component
+    # 1 rd_port: number of read port is 0.
+    # 1 wr_port: number of write port is 0.
+    #                         Type,     nregs,          rd_ports, wr_ports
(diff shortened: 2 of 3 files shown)


lib/messages.py
```diff
@@ -392,6 +392,31 @@ def str_func(s):
     namespace = {'__str__': str_func}
   )
 
+#=========================================================================
+# Ring for delivering data across SPM banks
```
tancheng, line 396: Ring is the NoC for multiple CGRAs? Or for multiple tiles?

controller/ControllerRTL.py
```diff
@@ -29,8 +29,14 @@ def construct(s, ControllerIdType, CmdType, CtrlPktType, NocPktType,
...
     s.send_to_noc = SendIfcRTL(NocPktType)
 
-    s.recv_from_cpu_ctrl_pkt = RecvIfcRTL(CtrlPktType)
-    s.send_to_ctrl_ring_ctrl_pkt = SendIfcRTL(CtrlPktType)
+    # s.recv_from_cpu_ctrl_pkt = RecvIfcRTL(CtrlPktType)
+    # s.send_to_ctrl_ring_ctrl_pkt = SendIfcRTL(CtrlPktType)
+    s.recv_ctrl_pkt = RecvIfcRTL(CtrlPktType)
+    s.send_ctrl_pkt = SendIfcRTL(CtrlPktType)
+    s.send_to_local_ctrl_ring_ctrl_pkt = SendIfcRTL(CtrlPktType)
+    s.recv_data_pkt = RecvIfcRTL(SPMDataPktType)
```
tancheng, line 37: What is this used for? Is it like CPU issuing some data that needs to be updated into certain SPM banks?

controller/ControllerRTL.py
```diff
@@ -29,8 +29,14 @@ def construct(s, ControllerIdType, CmdType, CtrlPktType, NocPktType,
     s.recv_from_noc = RecvIfcRTL(NocPktType)
     s.send_to_noc = SendIfcRTL(NocPktType)
 
-    s.recv_from_cpu_ctrl_pkt = RecvIfcRTL(CtrlPktType)
-    s.send_to_ctrl_ring_ctrl_pkt = SendIfcRTL(CtrlPktType)
+    # s.recv_from_cpu_ctrl_pkt = RecvIfcRTL(CtrlPktType)
+    # s.send_to_ctrl_ring_ctrl_pkt = SendIfcRTL(CtrlPktType)
+    s.recv_ctrl_pkt = RecvIfcRTL(CtrlPktType)
```
tancheng, line 34: Why do we remove `_cpu_` here? It can come from other places?

tancheng:
> example

Why do we remove most of the ifcs? How to handle different types if they are removed.

controller/ControllerRTL.py
yyan7223: So you want the data preloaded to SPM banks are also included in `recv_from_cpu_ctrl_pkt`?
Then I think the naming of this signal should be replaced with `recv_from_cpu_preloading_pkt`

controller/ControllerRTL.py
yyan7223: If we have 100 cgras, I don't think `let CPU connect all of them` is a good idea. 
Instead, only the first cgra connect to cpu, and the rest of cgras are connected sequentially, which will be more synthesised&P&R friendly I believe?
Or I change it back.

lib/messages.py
yyan7223: I mean the multiple banks in the SPM of one CGRA.

yyan7223:
> > example
> 
> Why do we remove most of the ifcs? How to handle different types if they are removed.

lib/messages.py
tancheng: Then all the banks are connected to one controller: https://github.com/tancheng/VectorCGRA/blob/4688d88207feb646a3574ce504880ede3dc40a94/cgra/CgraRTL.py#L89-L91

Why do we need to send pkt/msg on ring for SRAM banks?

controller/ControllerRTL.py
tancheng: I think `recv_from_cpu_pkt` could be better?

lib/messages.py
tancheng: Ah, do you mean the SRAM banks on other CGRAs?

controller/ControllerRTL.py
tancheng: Yes, we are aligned on this, CPU only talks to the first controller. Thus, we already have:
 - pkg from NoC: https://github.com/tancheng/VectorCGRA/blob/4688d88207feb646a3574ce504880ede3dc40a94/controller/ControllerRTL.py#L29
 - Handler (TODO) upon receiving the pkg from NoC: https://github.com/tancheng/VectorCGRA/blob/4688d88207feb646a3574ce504880ede3dc40a94/controller/ControllerRTL.py#L215

lib/messages.py
yyan7223: No, just local SRAM banks for current CGRA.
This is for preloading data to SRAM banks before executing, just like we load ctrl signals to tiles

controller/ControllerRTL.py
yyan7223: Yes it is.

controller/ControllerRTL.py
yyan7223: I don't understand, then what is the purpose of `recv_from_cpu_ctrl_pkt`?

yyan7223: Don't worry, let me read `CgraTemplateRTL.py` and understand your design.

tancheng: Thanks. And to briefly clarify, `recv_from_cpu_ctrl_pkt` of the first CGRA/controller would get pkt from CPU then send into the queue `recv_ctrl_pkt_queue`, then the pkt in the queue would be popped to the inter-CGRA NoC (now is a ring), note that even the CGRA_0's pkt would be sent into the ring and popped, to be uniformed/simplified.

lib/messages.py
tancheng: Would it be better if this message just have `data` and `addr`? The cgra id and bank id could be inferred based on the `addr2controller_lut`, and send through https://github.com/tancheng/VectorCGRA/blob/4688d88207feb646a3574ce504880ede3dc40a94/controller/ControllerRTL.py#L52-L53, which are already connected to the write ports of the data memory.

tancheng: I draw the arch figure for you:

image

PR #64, opened 2025-01-11
Add ConstQueueDynamicRTL

yuqisun:
Hi @tancheng ,

Could you help take a look if anything wrong with the RTL or RTL_test code, got below PyMTL error

```
E   pymtl3.dsl.errors.NoWriterError: The following nets need drivers.
E   Net:
E    - s.src_const_pkt.send.rdy
E    - s.const_queue.recv_const.rdy
```

Thanks,

### DIFF
--- a/mem/const/ConstQueueDynamicRTL.py
+++ b/mem/const/ConstQueueDynamicRTL.py
@@ -2,7 +2,8 @@
 ==========================================================================
 ConstQueueDynamicRTL.py
 ==========================================================================
-Constant memory with regs used for simulation.
+Constant Queue with regs used for simulation.
+If queue is full, will stop receiving new data.
 
 Author : Yuqi Sun
   Date : Jan 11, 2025
@@ -23,9 +24,9 @@ class ConstQueueDynamicRTL(Component):
     # ConstMemAddrType is 3 bits
     # ConstMemAddrTyp(1) = 001
     # ConstMemAddrType(2) = 010
-    AddrType = mk_bits(clog2(const_mem_size))
+    AddrType = mk_bits(max(1, clog2(const_mem_size)))
 
-    # cur to record the current valid address
+    # cur to record the current valid address, max is const_mem_size - 1
     s.cur = Wire(AddrType)
 
     # Interface
@@ -41,23 +42,33 @@ class ConstQueueDynamicRTL(Component):
     # Connections
     s.send_const.msg //= s.reg_file.rdata[0]
 
-    @update_ff
-    def write_to_reg():
-      s.recv_const.rdy <<= 1
+
+    @update
+    def update_msg():
+      s.recv_const.rdy @= 1
       # check if there's a valid const to be written
       if s.recv_const.val:
-        # .wen: enable write
-        # wen=1 to enable write, port 0
-        # data can be written only if wen
-        s.reg_file.wen[0] <<= 1
-        # if cur point to last element in mem
-        # drop data for now
-        if s.cur == AddrType(const_mem_size - 1):
-          print(f"Drop data as full const mem, size: {const_mem_size}")
-        else:
+        s.reg_file.waddr[0] @= s.cur
+        s.reg_file.wdata[0] @= s.recv_const.msg
+        s.reg_file.wen[0] @= 1
+      # s.recv_const.rdy @= 0 will stop receive const from inport immediately
+      # so will avoid receiving new data when regs full
+      if s.cur == AddrType(const_mem_size - 1):
+          s.recv_const.rdy @= 0
+
+
+    @update_ff
+    def write_to_reg():
+      # move cur in @update_ff
+      # if producer val and consumer(self) rdy
+      if s.recv_const.val & s.recv_const.rdy:
+        # move cur only if cur lt const_mem_size - 1
+        # and cur will plug 1 in the last loop
+        # then will update the last element in @update
+        # after the last element insert, will not move cur and receive const from inport(as set rdy 0 in @update)
+        if s.cur < AddrType(const_mem_size - 1):
           s.cur <<= s.cur + AddrType(1)
-          s.reg_file.waddr[0] <<= s.cur
-          s.reg_file.wdata[0] <<= s.recv_const.msg
+
 
     @update_ff
     def update_raddr():
@@ -71,5 +82,5 @@ class ConstQueueDynamicRTL(Component):
 
   def line_trace(s):
     const_mem_str  = "|".join([str(data) for data in s.reg_file.regs])
-    return f'const_mem_str: {const_mem_str}'
+    return f'\nconst_mem_str: {const_mem_str}'
 
--- a/mem/const/ConstQueueDynamicRTL.py
+++ b/mem/const/ConstQueueDynamicRTL.py
@@ -27,8 +27,12 @@ class ConstQueueDynamicRTL(Component):
     # ConstMemAddrType(2) = 010
     AddrType = mk_bits(max(1, clog2(const_mem_size)))
 
+    # Makes write cursor type 1 bit more than mem addr type as need compare with const_mem_size,
+    # otherwise, number will be back to 000 when 111 + 1 (given const_mem_size = 8)
+    WrCurType = mk_bits(clog2(const_mem_size + 1))
+
     # write cursor and read cursor
-    s.wr_cur = Wire(AddrType)
+    s.wr_cur = Wire(WrCurType)
     s.rd_cur = Wire(AddrType)
 
     # Interface
@@ -48,8 +52,9 @@ class ConstQueueDynamicRTL(Component):
 
     @update
     def load_const():
-      # Checks if full with s.recv_const.rdy, it comes from not_full in 'update_wr_cur()'.
-      if s.recv_const.val & s.recv_const.rdy:
+      not_full = s.wr_cur < const_mem_size
+      s.recv_const.rdy @= not_full
+      if s.recv_const.val & not_full:
         s.reg_file.waddr[0] @= trunc(s.wr_cur, AddrType)
         s.reg_file.wdata[0] @= s.recv_const.msg
         s.reg_file.wen[0] @= 1
@@ -57,17 +62,16 @@ class ConstQueueDynamicRTL(Component):
 
     @update_ff
     def update_wr_cur():
-      not_full = (s.wr_cur < (const_mem_size - 1))
-      s.recv_const.rdy <<= not_full
-      # Checks if there's a valid const(producer) to be written.
+      not_full = (s.wr_cur < const_mem_size)
+      # Checks if there's a valid const(from producer) to be written.
       if s.recv_const.val & not_full:
         s.wr_cur <<= s.wr_cur + 1
 
 
     @update
     def update_send_val():
-      # Checks if read cursor is in front of write cursor or both move to the last element.
-      if ((s.rd_cur < s.wr_cur) | (s.rd_cur == const_mem_size - 1)) & (s.wr_cur > 0):
+      # Checks if read cursor is in front of write cursor and regs is not empty.
+      if (zext(s.rd_cur, WrCurType) < s.wr_cur) & (s.wr_cur > 0):
         s.send_const.val @= 1
       else:
         s.send_const.val @= 0
@@ -77,7 +81,7 @@ class ConstQueueDynamicRTL(Component):
     def update_rd_cur():
       # Checks remote rdy.
       if s.send_const.rdy:
-        if s.rd_cur < s.wr_cur:
+        if zext((s.rd_cur), WrCurType) < s.wr_cur:
           s.rd_cur <<= s.rd_cur + 1
         else:
           s.rd_cur <<= 0
@@ -103,5 +107,6 @@ class ConstQueueDynamicRTL(Component):
       }
       reg_list.append(reg_dict)
     res_md = markdown_table(reg_list).set_params(quote = False).get_markdown()
-    return f"send.val: {self.send_const.val}{res_md}"
+    return (f"wr_cur: {self.wr_cur}, rd_cur: {self.rd_cur}, send.val: {self.send_const.val}, send_const.rdy: {self.send_const.rdy}"
+            f"{res_md}")
 
--- a/mem/const/ConstQueueDynamicRTL.py
+++ b/mem/const/ConstQueueDynamicRTL.py
@@ -56,7 +56,7 @@ class ConstQueueDynamicRTL(Component):
       # so will avoid receiving new data when regs full
       # can NOT do this by comparing cur > AddrType(const_mem_size - 1)
       # for example: mem size=8, cur will be Bits3(0x0) once it reaches Bits3(0x7) + AddrType(1) in loop
-      # and can NOT use bigger AddrType for cur as it will be used to update waddr: (s.reg_file.waddr[0] @= s.cur), waddr is binded with mem size
+      # and can NOT use bigger AddrType such as Bits4 for cur as it will be used to update waddr: (s.reg_file.waddr[0] @= s.cur), waddr is binded with mem size which will be Bits3 for mem size 8
       if s.cur == AddrType(const_mem_size - 1):
           s.recv_const.rdy @= 0
 
--- a/mem/const/ConstQueueDynamicRTL.py
+++ b/mem/const/ConstQueueDynamicRTL.py
@@ -51,7 +51,7 @@ class ConstQueueDynamicRTL(Component):
         s.reg_file.waddr[0] @= s.cur
         s.reg_file.wdata[0] @= s.recv_const.msg
         s.reg_file.wen[0] @= 1
-      # s.recv_const.rdy @= 0 will stop receive const from inport immediately
+      # s.recv_const.rdy @= 0 will stop receive const from inport immediately even there's rdy @= 1 in @update
       # so will avoid receiving new data when regs full
       if s.cur == AddrType(const_mem_size - 1):
           s.recv_const.rdy @= 0
--- a/mem/const/ConstQueueDynamicRTL.py
+++ b/mem/const/ConstQueueDynamicRTL.py
@@ -78,10 +78,16 @@ class ConstQueueDynamicRTL(Component):
       # check remote rdy and self val(val = 1 when all const saves to regs)
       if s.send_const.rdy & s.send_const.val:
         # read to the last element in mem, reset to addr to read from addr 0
-        if s.reg_file.raddr[0] == s.cur:
-          s.reg_file.raddr[0] <<= AddrType(0)
-        else:
-          s.reg_file.raddr[0] <<= s.reg_file.raddr[0] + AddrType(1)
+        # if s.reg_file.raddr[0] == s.cur:
+        #   s.reg_file.raddr[0] <<= AddrType(0)
+        # else:
+        #   s.reg_file.raddr[0] <<= s.reg_file.raddr[0] + AddrType(1)
+
+        # will not reset addr to 0 when read to the last element in mem
+        # as this solution read/write simultaneously
+        # cur always equals to raddr[0] + 1
+        # if wants to change to reset mode, can make read start after write finish
+        s.reg_file.raddr[0] <<= s.reg_file.raddr[0] + AddrType(1)
 
 
   def line_trace(s):
--- a/mem/const/ConstQueueDynamicRTL.py
+++ b/mem/const/ConstQueueDynamicRTL.py
@@ -27,7 +27,7 @@ class ConstQueueDynamicRTL(Component):
     # ConstMemAddrType(2) = 010
     AddrType = mk_bits(max(1, clog2(const_mem_size)))
 
-    # make Write cursor type 1 bit more than mem addr type as need compare if cursor > AddrType(const_mem_size - 1)
+    # Makes write cursor type 1 bit more than mem addr type as need compare if cursor > AddrType(const_mem_size - 1)
     # otherwise, number will be back to 000 when 111 + 1
     WrCurType = mk_bits(clog2(const_mem_size + 1))
 
@@ -42,7 +42,7 @@ class ConstQueueDynamicRTL(Component):
     # Component
     # 1 rd_port: number of read port is 0
     # 1 wr_port: number of write port is 0
-    #                               Type,     nregs,          rd_ports, wr_ports
+    #                         Type,     nregs,          rd_ports, wr_ports
     s.reg_file = RegisterFile(DataType, const_mem_size, 1,        1)
 
     # Connections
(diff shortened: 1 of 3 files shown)


mem/const/ConstQueueDynamicRTL.py
```diff
@@ -0,0 +1,74 @@
...
+    # ConstMemAddrTyp(1) = 001
+    # ConstMemAddrType(2) = 010
+    AddrType = mk_bits(clog2(const_mem_size))
+
+    # cur to record the current valid address
+    s.cur = Wire(AddrType)
+
+    # Interface
+    s.send_const = SendIfcRTL(DataType)
+    s.recv_const = RecvIfcRTL(DataType)
```
tancheng, line 33: I suggest to have a [NormalQueueRTL](https://github.com/tancheng/VectorCGRA/blob/cbd277b2062a26181f068bb182de74ea6972753c/mem/ctrl/CtrlMemDynamicRTL.py#L53) in this module to make the timing better and isolate the hand shakes. Anyways, I believe this is your initial prototyping.

mem/const/ConstQueueDynamicRTL.py
```diff
@@ -0,0 +1,74 @@
...
+    # 1 wr_port: number of write port is 0
+    s.reg_file = RegisterFile(DataType, const_mem_size, 1,        1)
+
+    # Connections
+    s.send_const.msg //= s.reg_file.rdata[0]
+
+    @update_ff
+    def write_to_reg():
+      # check if there's a valid const to be written
+      if s.recv_const.val == 1:
```
tancheng, line 47: Plz use `if s.recv_const.val:`, the `== 1` might work or not, I am not sure, as the `1` might be represented in different number of bits after translation from python to Verilog. (instead, the `.val` is 1 bit.) though pymtl might be smart enough to infer the number of bits of the `1`.

mem/const/ConstQueueDynamicRTL.py
```diff
@@ -0,0 +1,74 @@
...
+        # .wen: enable write
+        # wen=1 to enable write, port 0
+        # data can be written only if wen
+        s.reg_file.wen[0] <<= 1
```
tancheng, line 51: Once u have a `NormalQueueRTL`, you can set `wen` as `1` only when the input data is valid: https://github.com/tancheng/VectorCGRA/blob/cbd277b2062a26181f068bb182de74ea6972753c/mem/ctrl/CtrlMemDynamicRTL.py#L73

Makes sense? If there is no input valid, why we write sth into registers? When you set `wen` as `1`, it must write sth, so if no input valid, it probably writes some stale trash data into certain addr.

tancheng:
> Hi @tancheng ,
> 
> Could you help take a look if anything wrong with the RTL or RTL_test code, got below PyMTL error
> 
> ```
> E   pymtl3.dsl.errors.NoWriterError: The following nets need drivers.
> E   Net:
> E    - s.src_const_pkt.send.rdy
> E    - s.const_queue.recv_const.rdy
> ```
> 
> Thanks,

This means no one set `const_queue.recv_const.rdy` as `1`, it never reaches `1`. If you don't plan to include a `NormalQueueRTL` for now, you can set it as `1` within any of your `update` or `update_ff` blocks, indicating the consumer is guaranteed to be always ready to receive a const data (I guess this is indeed true in our case).

yuqisun:
It works!
Tested with write and read operation and added my understanding. Could you help review, thanks.

mem/const/ConstQueueDynamicRTL.py
yuqisun: Yes, think so.

mem/const/ConstQueueDynamicRTL.py
yuqisun: done

mem/const/ConstQueueDynamicRTL.py
yuqisun: done

mem/const/ConstQueueDynamicRTL.py
```diff
@@ -0,0 +1,103 @@
...
+        #   s.reg_file.raddr[0] <<= AddrType(0)
+        # else:
+        #   s.reg_file.raddr[0] <<= s.reg_file.raddr[0] + AddrType(1)
+
+        # Approach 2:
+        # will NOT reset addr to 0 when read to the last element in mem
+        # as this approach read/write simultaneously
+        # raddr[0] always equals to cur - 1
+        # if wants to change to reset mode, can make read start after write finish
+        s.reg_file.raddr[0] <<= s.reg_file.raddr[0] + AddrType(1)
```
tancheng, line 97: This is fine but a `rd_cur` might be better for readability. i.e., outside update blocks, do the connection:
```
s.reg_file.raddr[0] //= s.rd_cur
```
Then here:
```
@update_ff
def update_rd_cur():
  if s.send_const.rdy:
    if zext((s.rd_cur + 1), WrCurType) < s.wr_cur:
      s.rd_cur <<= s.rd_cur + 1 # AddrType(1) might not be necessary as the type of 1 can be inferred.
    else:
      s.rd_cur <<= 0
```
For now, we can assume 8 is for both `mem_size` and `required_size`. And we are assuming the read is slower than filling data into this `ConstQueue`. Later on, we do need to indicate (via an [InPort](https://github.com/tancheng/VectorCGRA/blob/cbd277b2062a26181f068bb182de74ea6972753c/noc/CrossbarRTL.py#L29)) the `required_size` (say 6, less than 8) is reached:
```
s.reg_file.raddr[0] //= s.rd_cur

s.wr_done = InPort(1) # set as `0` or `1` by another module (a parent module).

@update_ff
def update_rd_cur():
  if s.send_const.rdy & (zext(s.rd_cur, WrCurType) < s.wr_cur):
    if !s.wr_done | (zext((s.rd_cur + 1), WrCurType) < s.wr_cur):
      s.rd_cur <<= s.rd_cur + 1 # AddrType(1) might not be necessary as the type of 1 can be inferred.
    else: # condition of "s.wr_done & (zext((s.rd_cur + 1), WrCurType) >= s.wr_cur)"
      s.rd_cur <<= 0
```

mem/const/ConstQueueDynamicRTL.py
```diff
@@ -0,0 +1,103 @@
...
+        s.reg_file.waddr[0] @= s.cur
+        s.reg_file.wdata[0] @= s.recv_const.msg
+        s.reg_file.wen[0] @= 1
+      # s.recv_const.rdy @= 0(consumer: self) will stop receive const from inport immediately even there's rdy @= 1(self) in @update
+      # so will avoid receiving new data when regs full
+      # can NOT do this by comparing cur > AddrType(const_mem_size - 1)
+      # for example: mem size=8, cur will be Bits3(0x0) once it reaches Bits3(0x7) + AddrType(1) in loop
+      # and can NOT use bigger AddrType such as Bits4 for cur as it will be used to update waddr: (s.reg_file.waddr[0] @= s.cur), waddr is binded with mem size which will be Bits3 for mem size 8
+      if s.cur == AddrType(const_mem_size - 1):
+          s.recv_const.rdy @= 0
```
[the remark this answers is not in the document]
yuqisun: Tried same thing, got failed because reg_file.waddr[0] uses same AddrType bits with cur, and it related with reg size, if expand cur bits alone, it will not match in `s.reg_file.waddr[0] @= s.cur`.

Maybe cut the first bit when assign addr to reg_file.waddr would be a good idea?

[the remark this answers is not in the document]
tancheng: `Trunc(s.cur, AddrType)` would work.

yuqisun:
Hi Cheng,

Updated as discussed and added verbose_trace as below, please help review (only ConstQueueDynamicRTL.py and its test py and workflow would be good, others are merged from the latest master).

Screenshot from 2025-01-16 00-30-12

Thanks,

tancheng:
Instead of:
[yuqisun](https://github.com/yuqisun) wants to merge 22 commits into [tancheng:ConstQueueDynamicRTL](https://github.com/tancheng/VectorCGRA/tree/ConstQueueDynamicRTL) from [yuqisun:ConstMemRTL](https://github.com/yuqisun/VectorCGRA/tree/ConstMemRTL)
Shouldn't we have:
[yuqisun](https://github.com/yuqisun) wants to merge 22 commits into `master` from [yuqisun:ConstMemRTL](https://github.com/yuqisun/VectorCGRA/tree/ConstMemRTL)?
So there won't `19 File Changes`?

 - The markdown looks great.
 - Do you plan to add `delivering const via ring` in this PR?

mem/const/ConstQueueDynamicRTL.py
```diff
@@ -0,0 +1,112 @@
...
+      if s.recv_const.val & not_full:
+        s.reg_file.waddr[0] @= trunc(s.wr_cur, AddrType)
+        s.reg_file.wdata[0] @= s.recv_const.msg
+        s.reg_file.wen[0] @= 1
+
+
+    @update_ff
+    def update_wr_cur():
+      not_full = (s.wr_cur < (const_mem_size - 1))
+      # check if there's a valid const(producer) to be written
```
tancheng, line 67: move above `not_full`

mem/const/ConstQueueDynamicRTL.py
```diff
@@ -0,0 +1,112 @@
...
+      # have to add bracket if there's & and compare, i.e. s.recv_const.val & (s.wr_cur < const_mem_size)
```
tancheng, line 68: remove

mem/const/ConstQueueDynamicRTL.py
```diff
@@ -0,0 +1,112 @@
...
+    s.reg_file = RegisterFile(DataType, const_mem_size, 1,        1)
+
+    # Connections
+    s.send_const.msg //= s.reg_file.rdata[0]
+    s.reg_file.raddr[0] //= s.rd_cur
+
+
+    @update
+    def load_const():
+      not_full = s.wr_cur < const_mem_size
```
tancheng, line 55: add `()`

mem/const/ConstQueueDynamicRTL.py
tancheng: check -> Checks

mem/const/ConstQueueDynamicRTL.py
```diff
@@ -0,0 +1,112 @@
...
+    s.rd_cur = Wire(AddrType)
+
+    # Interface
+    s.send_const = SendIfcRTL(DataType)
+    s.recv_const = RecvIfcRTL(DataType)
+
+    # Component
+    # 1 rd_port: number of read port is 0
+    # 1 wr_port: number of write port is 0
+    #                               Type,     nregs,          rd_ports, wr_ports
```
tancheng, line 45: Align `Type` with `DataType`, etc

mem/const/ConstQueueDynamicRTL.py
```diff
@@ -0,0 +1,112 @@
...
+    # Constant
+    # addr type: number of bits to represent the address
+    # 2^addr_size = const_mem_size
+    # i.e. const_mem_size = 8
+    # ConstMemAddrType is 3 bits
+    # ConstMemAddrTyp(1) = 001
+    # ConstMemAddrType(2) = 010
+    AddrType = mk_bits(max(1, clog2(const_mem_size)))
+
+    # make Write cursor type 1 bit more than mem addr type as need compare if cursor > AddrType(const_mem_size - 1)
```
tancheng, line 30: make -> Makes

tancheng: Write -> write

mem/const/ConstQueueDynamicRTL.py
```diff
@@ -0,0 +1,112 @@
...
+    @update
+    def update_send_val():
+      # rd_cur cannot be larger than wr_cur and there IS const in regs
+      if (zext(s.rd_cur, WrCurType) <= s.wr_cur) & (s.wr_cur > 0):
+        s.send_const.val @= 1
+
+
+    @update_ff
+    def update_rd_cur():
+      # check remote rdy
```
tancheng, line 82: check -> Checks

mem/const/ConstQueueDynamicRTL.py
```diff
@@ -0,0 +1,112 @@
...
+      # check if there's a valid const(producer) to be written
+      # have to add bracket if there's & and compare, i.e. s.recv_const.val & (s.wr_cur < const_mem_size)
+      if s.recv_const.val & not_full:
+        s.wr_cur <<= s.wr_cur + 1
+
+
+    @update
+    def update_send_val():
+      # rd_cur cannot be larger than wr_cur and there IS const in regs
+      if (zext(s.rd_cur, WrCurType) <= s.wr_cur) & (s.wr_cur > 0):
```
tancheng, line 76: I doubt about `<=`? Shouldn't it be `<`? For example, at cycle `12`, `wr_cur` is `12`, and `rd_cur` is `12` as well, but `reg[12]` is not yet written (`reg[11]` is just written as write takes 1 cycle). WDYT?

mem/const/ConstQueueDynamicRTL.py
```diff
@@ -0,0 +1,112 @@
...
+        s.send_const.val @= 1
+
+
+    @update_ff
+    def update_rd_cur():
+      # check remote rdy
+      if s.send_const.rdy:
+        if zext((s.rd_cur), WrCurType) < s.wr_cur:
+          # type of 1 can be inferred to AddrType(1)
```
tancheng, line 85: remove

mem/const/test/ConstQueueDynamicRTL_test.py
```diff
@@ -0,0 +1,78 @@
...
+  test_harness.sim_tick()
+  print("\nmem: {}".format(test_harness.line_trace()))
+
+
+def test_const_num_lt_mem():
+  MemUnit = ConstQueueDynamicRTL
+  DataType = mk_data(4, 1)
+  const_mem_size = 8
+
+  # number of source const less than mem size
```
tancheng, line 57: `number of source const less than mem size` -> `The number of source const is less than the memory size.`

mem/const/ConstQueueDynamicRTL.py
yuqisun: `# check if there's a valid const(producer) to be written` this is comment for `if s.recv_const.val`

mem/const/ConstQueueDynamicRTL.py
yuqisun: Got it and add condition logic to make send.val 0

yuqisun:
> Instead of: [yuqisun](https://github.com/yuqisun) wants to merge 22 commits into [tancheng:ConstQueueDynamicRTL](https://github.com/tancheng/VectorCGRA/tree/ConstQueueDynamicRTL) from [yuqisun:ConstMemRTL](https://github.com/yuqisun/VectorCGRA/tree/ConstMemRTL) Shouldn't we have: [yuqisun](https://github.com/yuqisun) wants to merge 22 commits into `master` from [yuqisun:ConstMemRTL](https://github.com/yuqisun/VectorCGRA/tree/ConstMemRTL)? So there won't `19 File Changes`?
> 
>     * The markdown looks great.
> 
>     * Do you plan to add `delivering const via ring` in this PR?

Yes, let me start next step.

mem/const/ConstQueueDynamicRTL.py
tancheng: I just meant adding `s` on verbs and use capitalize the first char.

mem/const/ConstQueueDynamicRTL.py
```diff
@@ -0,0 +1,112 @@
...
+      # (rd_cur < wr_cur or read last const in mem) and wr_cur > 0
+      if ((zext(s.rd_cur, WrCurType) < s.wr_cur) | (s.rd_cur == const_mem_size - 1)) & (s.wr_cur > 0):
+        s.send_const.val @= 1
+      else:
+        s.send_const.val @= 0
+
+
+    @update_ff
+    def update_rd_cur():
+      # Checks remote rdy
```
tancheng, line 83: Lack of period at the end.

mem/const/ConstQueueDynamicRTL.py
```diff
@@ -0,0 +1,112 @@
...
+      if s.recv_const.val & not_full:
+        s.reg_file.waddr[0] @= trunc(s.wr_cur, AddrType)
+        s.reg_file.wdata[0] @= s.recv_const.msg
+        s.reg_file.wen[0] @= 1
+
+
+    @update_ff
+    def update_wr_cur():
+      not_full = (s.wr_cur < (const_mem_size - 1))
+      # Checks if there's a valid const(producer) to be written
```
tancheng, line 67: Lack of period at the end.

mem/const/ConstQueueDynamicRTL.py
```diff
@@ -0,0 +1,112 @@
...
+      if s.recv_const.val & not_full:
+        s.wr_cur <<= s.wr_cur + 1
+
+
+    @update
+    def update_send_val():
+      # (rd_cur < wr_cur or read last const in mem) and wr_cur > 0
```
tancheng, line 74: Change this to comments.

mem/const/ConstQueueDynamicRTL.py
```diff
@@ -0,0 +1,112 @@
...
+
+    # Connections
+    s.send_const.msg //= s.reg_file.rdata[0]
+    s.reg_file.raddr[0] //= s.rd_cur
+
+
+    @update
+    def load_const():
+      not_full = (s.wr_cur < const_mem_size)
+      # Checks if there's a valid const(from producer) to be written
```
tancheng, line 56: Lack of period at the end.

mem/const/ConstQueueDynamicRTL.py
```diff
@@ -0,0 +1,112 @@
...
+    # write cursor and read cursor
+    s.wr_cur = Wire(WrCurType)
+    s.rd_cur = Wire(AddrType)
+
+    # Interface
+    s.send_const = SendIfcRTL(DataType)
+    s.recv_const = RecvIfcRTL(DataType)
+
+    # Component
+    # 1 rd_port: number of read port is 0
```
tancheng, line 43: Lack of period at the end.

mem/const/ConstQueueDynamicRTL.py
```diff
@@ -0,0 +1,112 @@
...
+    # Constant
+    # addr type: number of bits to represent the address
+    # 2^addr_size = const_mem_size
+    # i.e. const_mem_size = 8
+    # ConstMemAddrType is 3 bits
+    # ConstMemAddrTyp(1) = 001
+    # ConstMemAddrType(2) = 010
+    AddrType = mk_bits(max(1, clog2(const_mem_size)))
+
+    # Makes write cursor type 1 bit more than mem addr type as need compare if cursor > AddrType(const_mem_size - 1)
```
tancheng, line 30: Lack of period at the end.

mem/const/ConstQueueDynamicRTL.py
```diff
@@ -0,0 +1,112 @@
...
+      not_full = (s.wr_cur < (const_mem_size - 1))
+      # Checks if there's a valid const(producer) to be written
+      if s.recv_const.val & not_full:
+        s.wr_cur <<= s.wr_cur + 1
+
+
+    @update
+    def update_send_val():
+      # (rd_cur < wr_cur or read last const in mem) and wr_cur > 0
+      if ((zext(s.rd_cur, WrCurType) < s.wr_cur) | (s.rd_cur == const_mem_size - 1)) & (s.wr_cur > 0):
```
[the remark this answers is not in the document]
yuqisun: I think wr_cur will not goes back to 0, s.wr_cur = (const_mem_size - 1) if full.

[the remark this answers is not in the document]
yuqisun: And we may not need WrCurType size = mem_size + 1 as there're both wr and rd cursor, right?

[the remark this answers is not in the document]
tancheng: Then we won't be able to figure out the case when both `wr_cur` and `rd_cur` are at `mem_size - 1`, wether the data is already written (`rd_cur` is faster). Do you think if we can make `wr_cur` be able to reach `mem_size` could solve this issue and make the logic easier? Can we try this?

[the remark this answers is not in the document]
yuqisun: Agree, revert to make wr_cur type bits as `mem_size + 1`.

mem/const/ConstQueueDynamicRTL.py
```diff
@@ -0,0 +1,112 @@
...
+      if s.recv_const.val & not_full:
+        s.reg_file.waddr[0] @= trunc(s.wr_cur, AddrType)
+        s.reg_file.wdata[0] @= s.recv_const.msg
+        s.reg_file.wen[0] @= 1
+
+
+    @update_ff
+    def update_wr_cur():
+      not_full = (s.wr_cur < const_mem_size)
+      # Checks if there's a valid const(from producer) to be written.
```
tancheng, line 66: `const(from` -> `const (from`

mem/const/ConstQueueDynamicRTL.py
```diff
@@ -0,0 +1,112 @@
...
+      if s.recv_const.val & not_full:
+        s.wr_cur <<= s.wr_cur + 1
+
+
+    @update
+    def update_send_val():
+      # Checks if read cursor is in front of write cursor and regs is not empty.
+      if (zext(s.rd_cur, WrCurType) < s.wr_cur) & (s.wr_cur > 0):
```
tancheng, line 74: `& (s.wr_cur > 0)` is necessary?

mem/const/ConstQueueDynamicRTL.py
```diff
@@ -0,0 +1,112 @@
...
+        s.send_const.val @= 1
+      else:
+        s.send_const.val @= 0
+
+
+    @update_ff
+    def update_rd_cur():
+      # Checks remote rdy.
```
tancheng, line 82: `# Checks remote rdy.` -> `# Checks whether the "reader" successfully read the data at rd_cur, and proceed rd_cur accordingly.` (in two lines)

tancheng (approved):
I would say let's merge this PR first then try to integrate this RTL into existing design (to make the review/comments more concise/relevant)?

mem/const/ConstQueueDynamicRTL.py
yuqisun: done

mem/const/ConstQueueDynamicRTL.py
yuqisun: You are right, removed it.

mem/const/ConstQueueDynamicRTL.py
yuqisun: Updated.

yuqisun:
Screenshot from 2025-01-19 15-17-05
Screenshot from 2025-01-19 15-18-28

Issue #39, opened 2024-12-11
[P2] Cleanup and update


tancheng:
Deprecate en/rdy and use val/rdy for all modules.

tancheng:
@RuihongY FYI.

tancheng:
Hi @yo96, assigning this cleanup bug to you.

tancheng:
The concrete TODOs:
 - [ ] Replace en/rdy with val/rdy
 - [ ] Resolve latches (provide initialization for all signals within a combinational block)
 - [ ] Update CGRA and PyOCN to latest PyMTL4.0 (or 3.0 whatever)
 - [ ] Make the xbar more modular rather than a behavioral-like design

tancheng:
Hi @yo96, would following introduce combinational loop?
```
s.send_out[0].msg.predicate @= s.send_out[0].msg.predicate
```
in https://github.com/tancheng/VectorCGRA/blob/49a40187f90a8d67b0a410c0d3bc5944422e22e6/fu/single/AdderRTL.py#L104C9-L104C67

yo96:
Yes, that looks like a combinational loop to me. It is like connecting the output of an AND gate to its input.

tancheng:
> The concrete TODOs:
> 
> * [ ]  Replace en/rdy with val/rdy
> * [ ]  Resolve latches (provide initialization for all signals within a combinational block)
> * [ ]  Update CGRA and PyOCN to latest PyMTL4.0 (or 3.0 whatever)
> * [ ]  Make the xbar more modular rather than a behavioral-like design

Updated:

 - [x]  Replace en/rdy with val/rdy: https://github.com/tancheng/VectorCGRA/pull/54
 - [ ]  Resolve latches (provide initialization for all signals within a combinational block)
 - [ ]  Update CGRA and PyOCN to latest PyMTL4.0 (or 3.0 whatever)
 - [ ]  Make the xbar more modular rather than a behavioral-like design

Issue #13, opened 2023-11-04
Hierarchical synthesis-friendly tile design


tancheng:
Existing tile design has combinational loop. Let's consider the option 2 towards a hierarchical synthesis-friendly tile design.

<img width="715" alt="tile" src="https://github.com/tancheng/VectorCGRA/assets/6756658/3794c84d-87e5-450c-8210-596740ed56f5">

yyan7223:
Hi cheng,

- Do you mean the combinationl loop in option 1 is `xbar -> regs -> FU -> xbar`, but we have a register in the loop that stablizes the input of xbar.
- I think there might be another issue in option 1. Assume the output data of `left FU` at cycle 1 will be sent to the input register of `right FU` at cycle 1 so that `right FU` can continuously process the data at cycle 2. Then the data should go through the path `left xbar -> link -> right xbar` during cycle 1, will that be too far for a high frequency like 1GHz?
- Do we still support bypass in option 2? There seems no red bypass lines for registers in the right of 8x8 xbar in option 2.

Regards,
Yufei

tancheng:
Many issues for Option 1, so I prefer Option 2. Bypass is only for bypassing the register on the right side of the 8x8 xbar, with Option 2, the outputs from FUs automatically bypass those registers.

tancheng:
Once we decoupled the FU outports control from the routing xbar, we can start to use DC compiler to resolve the potential combinational loops. @yyan7223

tancheng:
Note that there is also a `OrLink` between the `2x8 xbar outports` and the `FU registers inports` (didn't draw it).

yyan7223:
I cannot generate the verilog using CgraTemplateRTL_test.py, CgraRTL_test.py, and TileRTL_test.py. Could you please try on your side?
<img width="1244" alt="1736305504815" src="https://github.com/user-attachments/assets/f4fc7a05-25d5-452e-9293-795ec81dfcca" />

tancheng:
[...]

PR #65, opened 2025-01-11
Resolve combinational loops

tancheng:
The `val/rdy` ifcs are leveraged for hand-shake.
 - We allow `rdy` to be dependent on `val`, but should avoid `val` depending on `rdy`, otherwise, loops exist.

Related issues:
 - https://github.com/tancheng/VectorCGRA/issues/13
 - https://github.com/tancheng/VectorCGRA/issues/39

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -12,7 +12,7 @@ on:
 jobs:
   build:
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
       fail-fast: false
       matrix:
--- a/cgra/CgraRTL.py
+++ b/cgra/CgraRTL.py
@@ -72,6 +72,7 @@ class CgraRTL(Component):
                       data_mem_size_global, num_ctrl,
                       total_steps, 4, 2, s.num_mesh_ports,
                       s.num_mesh_ports,
+                      FuList = FuList,
                       const_list = preload_const[i])
               for i in range(s.num_tiles)]
     s.data_mem = DataMemWithCrossbarRTL(NocPktType, DataType,
--- a/cgra/CgraTemplateRTL.py
+++ b/cgra/CgraTemplateRTL.py
@@ -66,6 +66,7 @@ class CgraTemplateRTL(Component):
                       data_mem_size_global, num_ctrl,
                       total_steps, 4, 2, s.num_mesh_ports,
                       s.num_mesh_ports,
+                      FuList = FuList,
                       const_list = preload_const[i])
               for i in range(s.num_tiles)]
     # FIXME: Need to enrish data-SPM-related user-controlled parameters, e.g., number of banks.
--- a/cgra/test/CgraRTL_test.py
+++ b/cgra/test/CgraRTL_test.py
@@ -230,35 +230,35 @@ def test_homogeneous_2x2(cmdline_opts):
   th = config_model_with_cmdline_opts(th, cmdline_opts, duts = ['dut'])
   run_sim(th)
 
-# def test_heterogeneous_king_mesh_2x2(cmdline_opts):
-#   topology = "KingMesh"
-#   th = init_param(topology)
-#   th.set_param("top.dut.tile[1].construct", FuList=[ShifterRTL])
-#   th.elaborate()
-#   th.dut.set_metadata(VerilogVerilatorImportPass.vl_Wno_list,
-#                       ['UNSIGNED', 'UNOPTFLAT', 'WIDTH', 'WIDTHCONCAT',
-#                        'ALWCOMBORDER'])
-#   th = config_model_with_cmdline_opts(th, cmdline_opts, duts = ['dut'])
-#   run_sim(th)
-#
-# def test_vector_king_mesh_2x2(cmdline_opts):
-#   topology = "KingMesh"
-#   FuList = [AdderRTL,
-#             MulRTL,
-#             LogicRTL,
-#             ShifterRTL,
-#             PhiRTL,
-#             CompRTL,
-#             BranchRTL,
-#             MemUnitRTL,
-#             SelRTL,
-#             VectorMulComboRTL,
-#             VectorAdderComboRTL]
-#   th = init_param(topology, FuList)
-#   th.elaborate()
-#   th.dut.set_metadata(VerilogVerilatorImportPass.vl_Wno_list,
-#                       ['UNSIGNED', 'UNOPTFLAT', 'WIDTH', 'WIDTHCONCAT',
-#                        'ALWCOMBORDER'])
-#   th = config_model_with_cmdline_opts(th, cmdline_opts, duts = ['dut'])
-#   run_sim(th)
+def test_heterogeneous_king_mesh_2x2(cmdline_opts):
+  topology = "KingMesh"
+  th = init_param(topology)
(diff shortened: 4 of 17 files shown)


tancheng:
@yuqisun python 3.7 doesn't exist on latest ubuntu (https://github.com/actions/runner-images/issues/10636), seems like the new rule is starting being applied in our test workflow.

Do you think we need to keep ubuntu be 22.04, or deprecate python 3.7?

yuqisun:
> @yuqisun python 3.7 doesn't exist on latest ubuntu ([actions/runner-images#10636](https://github.com/actions/runner-images/issues/10636)), seems like the new rule is starting being applied in our test workflow.
> 
> Do you think we need to keep ubuntu be 22.04, or deprecate python 3.7?

May keep ubuntu-20.04 (same as CGRA-Mapper) until get LLVM upgraded? 
Tried in OS, not easy to install different version LLVM with default in OS, not sure how about in github workflow, can try later. If all good, we can deprecate py37, WDYT?

2 options to upgrade ubuntu:

1. Install LLVM-12 in new ubuntu (not easy in OS, you may try with mac also)
2. Upgrade LLVM version in code (sounds better?) (but remember there's arguments change between LLVM 12 and 18)

tancheng:
> > @yuqisun python 3.7 doesn't exist on latest ubuntu ([actions/runner-images#10636](https://github.com/actions/runner-images/issues/10636)), seems like the new rule is starting being applied in our test workflow.
> > Do you think we need to keep ubuntu be 22.04, or deprecate python 3.7?
> 
> May keep ubuntu-20.04 (same as CGRA-Mapper) until get LLVM upgraded? Tried in OS, not easy to install different version LLVM with default in OS, not sure how about in github workflow, can try later. If all good, we can deprecate py37, WDYT?
> 
> 2 options to upgrade ubuntu:
> 
> 1. Install LLVM-12 in new ubuntu (not easy in OS, you may try with mac also)
> 2. Upgrade LLVM version in code (sounds better?) (but remember there's arguments change between LLVM 12 and 18)
> 
> For now, we can use ubuntu-20.04 or disable py37 to pass build right?

So you mean for `ubuntu-22.04`, `LLVM-12` cannot be installed (I can't remember this, I thought any version of LLVM can be installed on any version of ubuntu...)? If that's the case, keeping `ubuntu-20.04` (instead of using `latest-ubuntu` in the action workflow) might be the easiest solution.

tancheng:
Hi @yyan7223, this PR may resolve some of the combinational loops issues. The combinational loops are identified by verilator itself via `verilator CgraRTL__xxx__pickled.v --lint-only`, e.g., `Feedback to clock or circular logic: xxx`.

Btw, @sgryin2@gmail.com (@yin00473@umn.edu) and @yo96 would also help on this by some synthesis tool. If you cannot get report from OpenRoad, plz consider DC compiler.

yuqisun:
> > > @yuqisun python 3.7 doesn't exist on latest ubuntu ([actions/runner-images#10636](https://github.com/actions/runner-images/issues/10636)), seems like the new rule is starting being applied in our test workflow.
> > > Do you think we need to keep ubuntu be 22.04, or deprecate python 3.7?
> > 
> > 
> > May keep ubuntu-20.04 (same as CGRA-Mapper) until get LLVM upgraded? Tried in OS, not easy to install different version LLVM with default in OS, not sure how about in github workflow, can try later. If all good, we can deprecate py37, WDYT?
> > 2 options to upgrade ubuntu:
> > 
> > 1. Install LLVM-12 in new ubuntu (not easy in OS, you may try with mac also)
> > 2. Upgrade LLVM version in code (sounds better?) (but remember there's arguments change between LLVM 12 and 18)
> > 
> > For now, we can use ubuntu-20.04 or disable py37 to pass build right?
> 
> So you mean for `ubuntu-22.04`, `LLVM-12` cannot be installed (I can't remember this, I thought any version of LLVM can be installed on any version of ubuntu...)? If that's the case, keeping `ubuntu-20.04` (instead of using `latest-ubuntu` in the action workflow) might be the easiest solution.

Yes, think so, no luck when I try to install LLVM-12 in ubuntu-24.04 OS (default 18).

tancheng:
> > > > @yuqisun python 3.7 doesn't exist on latest ubuntu ([actions/runner-images#10636](https://github.com/actions/runner-images/issues/10636)), seems like the new rule is starting being applied in our test workflow.
> > > > Do you think we need to keep ubuntu be 22.04, or deprecate python 3.7?
> > > 
> > > 
> > > May keep ubuntu-20.04 (same as CGRA-Mapper) until get LLVM upgraded? Tried in OS, not easy to install different version LLVM with default in OS, not sure how about in github workflow, can try later. If all good, we can deprecate py37, WDYT?
> > > 2 options to upgrade ubuntu:
> > > 
> > > 1. Install LLVM-12 in new ubuntu (not easy in OS, you may try with mac also)
> > > 2. Upgrade LLVM version in code (sounds better?) (but remember there's arguments change between LLVM 12 and 18)
> > > 
> > > For now, we can use ubuntu-20.04 or disable py37 to pass build right?
> > 
> > 
> > So you mean for `ubuntu-22.04`, `LLVM-12` cannot be installed (I can't remember this, I thought any version of LLVM can be installed on any version of ubuntu...)? If that's the case, keeping `ubuntu-20.04` (instead of using `latest-ubuntu` in the action workflow) might be the easiest solution.
> 
> Yes, think so, no luck when I try to install LLVM-12 in ubuntu-24.04 OS (default 18).

Thanks! No worry, I changed our default ubuntu version to 22.04 in the action workflow.

yyan7223:
Hi cheng @tancheng 

The attached file [compiledc.log](https://github.com/user-attachments/files/18394144/compiledc.log) is the combinational loop report from DC compiler. Most of loops are on `fu_crossbar` and `routing_crossbar`. However, only 20 loops are reported, which is much less then 150 loops reported by Openroad.

By the way, I test `verilator CgraRTL__xxx__pickled.v --lint-only` on my side using docker image provided by Yuqi,
verilator automatically exits due to too many warning(s) and I cannot see any `Feedback to clock or circular logic: xxx`


So does DC compiler provide more details than Verilator on reporting combinational loops? If that can help us better resolve the combinational loops, I can further investigate why only 20 loops are reported in DC compiler.

regards,
yufei

tancheng:
> The attached file [compiledc.log](https://github.com/user-attachments/files/18394144/compiledc.log) is the combinational loop report from DC compiler. Most of loops are on `fu_crossbar` and `routing_crossbar`. However, only 20 loops are reported, which is much less then 150 loops reported by Openroad.

Thanks Yufei! I saw all 20 loops are caused by `SelRTL` and I have updated/fixed in the latest commit, can you sync and try again? Hope other loops would be reported or the number of loops decreases.

> By the way, I test `verilator CgraRTL__xxx__pickled.v --lint-only` on my side using docker image provided by Yuqi, verilator automatically exits due to too many warning(s) and I cannot see any `Feedback to clock or circular logic: xxx` 

I have the exactly same msg as yours. And the `Feedback to clock or circular logic: xxx` appears when I scroll up my screen/terminal.

yyan7223 (approved):

yyan7223:
Cool, no loops reported by DC compiler now. @tancheng

tancheng:
Cool, then let me merge this PR :-)

PR #71, opened 2025-01-23
Add linetrace=True in DefaultPassGroup

yuqisun:
`test_harness.apply(DefaultPassGroup(linetrace = True))` can enable the simulator to automatically call the line_trace method.

### DIFF
--- a/mem/const/test/ConstQueueDynamicRTL_test.py
+++ b/mem/const/test/ConstQueueDynamicRTL_test.py
@@ -42,12 +42,9 @@ def run_sim(test_harness, max_cycles = 20):
   while not test_harness.done() and ncycles < max_cycles:
     test_harness.sim_tick()
     ncycles += 1
-    # print("\n{}: {}".format(ncycles, test_harness.line_trace()))
 
   for i in range(3):
     test_harness.sim_tick()
-    # print("\nextra clk {}: {}".format(i, test_harness.line_trace()))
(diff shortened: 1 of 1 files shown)


yuqisun:
Also find an interesting method which can print each value by clk, such as below:
Would it be helpful in verbose trace? (looks cool)
Screenshot from 2025-01-23 23-35-20

mem/const/test/ConstQueueDynamicRTL_test.py
```diff
@@ -35,19 +35,19 @@ def line_trace(s):
...
-  test_harness.apply(DefaultPassGroup())
+  test_harness.apply(DefaultPassGroup(linetrace = True))
   test_harness.sim_reset()
 
   ncycles = 0
   while not test_harness.done() and ncycles < max_cycles:
     test_harness.sim_tick()
     ncycles += 1
-    print("\n{}: {}".format(ncycles, test_harness.line_trace()))
+    # print("\n{}: {}".format(ncycles, test_harness.line_trace()))
```
tancheng, line 45: Would `tick 0` and `tick 1` be printed out with this default method?

Also, plz remove the commented out lines.

yuqisun: Yes, it prints from  `1r wr_cur: 0, rd_cur: 0, send.val: 0, send_const.rdy: 1`

tancheng (approved):

PR #77, opened 2025-02-04
Controller: preloading ctrl & data

yyan7223:


controller/ControllerRTL.py
```diff
@@ -249,6 +302,42 @@ def update_received_msg():
         #   # TODO: Handle other cmd types.
         #   assert(False)
 
+      if controller_id == 0:
```
tancheng, line 305: Why do we still have such check, this is not aligned with our discussion

controller/ControllerRTL.py
```diff
@@ -68,9 +68,9 @@ def construct(s, ControllerIdType, CmdType, CtrlPktType, NocPktType,
     # allow one request be sent out per cycle).
     # TODO: Include other cmd requests, e.g., dynamic rescheduling,
     # termination).
-    s.crossbar = XbarBypassQueueRTL(NocPktType, 3, 1)
+    s.crossbar = XbarBypassQueueRTL(NocPktType, 4, 1)
 
-    s.recv_ctrl_pkt_queue = NormalQueueRTL(CtrlPktType)
+    s.recv_ctrl_pkt_queue = ChannelRTL(FromCpuPktType, latency = 1)
```
tancheng, line 73: The `recv_ctrl_pkt_queue` should be renamed to `recv_from_cpu_pkt_queue`.

controller/ControllerRTL.py
yyan7223: Do you mean the input of `recv_ctrl_pkt_queue` is only `recv_from_cpu_pkt`,
and the decoded ctrl signals from `recv_from_noc` are directly sent to `send_to_ctrl_ring_ctrl_pkt`?

tancheng: Right, just isolate `CMD_CONFIG` from `elif (s.recv_from_noc.msg.cmd == CMD_STORE_REQUEST) | (s.recv_from_noc.msg.cmd == CMD_CONFIG)`, send the ctrl signals to the `ctrl_ring` here. How does it sound?

PR #81, opened 2025-02-17
[fix] Fix combinational loop in VectorAdderRTL.py

tancheng:
`val` wrongly depend on `rdy` in `VectorAdderRTL.py`.


tancheng:
Hi @RuihongY, PTAL and grant approval, thanks~!

yuqisun (approved):

PR #100, opened 2025-03-27
X-prop and inferred latch fix proposal.

rp15:
The reset issue was found by not being able to write values into the const memories from a custom SystemVerilog test bench. This led to observing the ready signal from the constant memory as always undefined (X). It could be traced back to the wr_cur DFF.

The other DFF's missing reset and the incomplete if-statement were found by subsequently reviewing the rest of the mem/const/ConstQueueDynamicRTL.py file.

We should think about reviewing the whole code base for similar potential issues.

### DIFF
--- a/mem/const/ConstQueueDynamicRTL.py
+++ b/mem/const/ConstQueueDynamicRTL.py
@@ -58,14 +58,21 @@ class ConstQueueDynamicRTL(Component):
         s.reg_file.waddr[0] @= trunc(s.wr_cur, AddrType)
         s.reg_file.wdata[0] @= s.recv_const.msg
         s.reg_file.wen[0] @= 1
+      else:
+        s.reg_file.waddr[0] @= trunc(s.wr_cur, AddrType)
+        s.reg_file.wdata[0] @= s.recv_const.msg
+        s.reg_file.wen[0] @= 0
 
 
     @update_ff
     def update_wr_cur():
       not_full = (s.wr_cur < const_mem_size)
+      if s.reset:
+        s.wr_cur <<= 0
       # Checks if there's a valid const (from producer) to be written.
-      if s.recv_const.val & not_full:
-        s.wr_cur <<= s.wr_cur + 1
+      else:
+        if s.recv_const.val & not_full:
+          s.wr_cur <<= s.wr_cur + 1
 
 
     @update
@@ -79,13 +86,16 @@ class ConstQueueDynamicRTL(Component):
 
     @update_ff
     def update_rd_cur():
+      if s.reset:
+        s.rd_cur <<= 0
       # Checks whether the "reader" successfully read the data at rd_cur,
       # and proceed rd_cur accordingly.
-      if s.send_const.rdy:
-        if zext((s.rd_cur), WrCurType) < (s.wr_cur - 1):
-          s.rd_cur <<= s.rd_cur + 1
-        else:
-          s.rd_cur <<= 0
+      else:
+        if s.send_const.rdy:
+          if zext((s.rd_cur), WrCurType) < (s.wr_cur - 1):
+            s.rd_cur <<= s.rd_cur + 1
+          else:
+            s.rd_cur <<= 0
 
 
   def line_trace(s, verbosity = 0):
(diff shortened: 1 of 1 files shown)


mem/const/ConstQueueDynamicRTL.py
```diff
@@ -58,14 +58,21 @@ def load_const():
         s.reg_file.waddr[0] @= trunc(s.wr_cur, AddrType)
         s.reg_file.wdata[0] @= s.recv_const.msg
         s.reg_file.wen[0] @= 1
+      else:
+        s.reg_file.waddr[0] @= trunc(s.wr_cur, AddrType)
+        s.reg_file.wdata[0] @= s.recv_const.msg
```
tancheng, line 63: Can we set them to `0` for `else` case?
```
        s.reg_file.waddr[0] @= 0
        s.reg_file.wdata[0] @= 0
```

rp15: As long as wen is deasserted, it should not matter what the value of those signals are. Keeping them unchanged  would save us two MUXes in theory. In practice, the synthesis tool will probably figure out that those signals are irrelevant when 0 == wen and optimize away the MUXes anyway.

My preferred way to write this would be to move waddr and wdata outside the if and only make wen conditional. But it is also OK to assign them in a way that best matches the coding style throughout the repo.

tancheng: > My preferred way to write this would be to move waddr and wdata outside the if and make only wen conditional. But it is also OK to assign them in a way that best matches the coding style throughout the repo.

This is also the style we are following, so can we change it to:
```
    @update
    def load_const():
      # Initializes signals.
      s.reg_file.waddr[0] @= 0
      s.reg_file.wdata[0] @= 0
      s.reg_file.wen[0] @= 1

      not_full = s.wr_cur < const_mem_size
      s.recv_const.rdy @= not_full

      if s.recv_const.val & not_full:
        s.reg_file.waddr[0] @= trunc(s.wr_cur, AddrType)
        s.reg_file.wdata[0] @= s.recv_const.msg
        s.reg_file.wen[0] @= 1
```
WDYT?

rp15: As long as we change the first wen assignment to 0, I am all for it. I would also double check the resulting .v RTL, but it looks good.

rp15: This is what that turned into (with the wen-init change), should be functionally correct!

```
   8367   always_comb begin : load_const
   8368     reg_file__waddr[1'd0] = 12'd0;
   8369     reg_file__wdata[1'd0] = 35'd0;
   8370     reg_file__wen[1'd0] = 1'd0;
   8371     __tmpvar__load_const_not_full = wr_cur < 13'( __const__const_mem_size_at_load_const );
   8372     recv_const__rdy = __tmpvar__load_const_not_full;
   8373     if ( recv_const__val & __tmpvar__load_const_not_full ) begin
   8374       reg_file__waddr[1'd0] = 12'(wr_cur);
   8375       reg_file__wdata[1'd0] = recv_const__msg;
   8376       reg_file__wen[1'd0] = 1'd1;
   8377     end
   8378   end
```

tancheng: Cool, plz make corresponding changes so we can merge, thanks!

rp15: I think we have to have 0 casted to the correct data types for wdata and waddr initialization, plain 0 is failing. Do you know how to do that?

mem/const/ConstQueueDynamicRTL.py
```diff
@@ -52,16 +52,18 @@ def construct(s, DataType, const_mem_size):
 
     @update
     def load_const():
+      # Initializes signals.
+      s.reg_file.waddr[0] @= 0
```
tancheng, line 56: Try to change `0` to `AddrType()`?

mem/const/ConstQueueDynamicRTL.py
```diff
@@ -52,16 +52,18 @@ def construct(s, DataType, const_mem_size):
...
+      s.reg_file.wdata[0] @= 0
```
tancheng, line 57: Try to change `0` to `DataType()`?

tancheng (approved):
LGTM

Issue #105, opened 2025-03-30
[P0] Add missing resets to four flopped signals in CtrlMemDynamicRTL.py


rp15:
s.start_iterate_ctrl
s.sent_complete
s.times
s.reg_file.raddr[0]

Issue #73, opened 2025-02-02
[P1] First round/iteration execution shouldn't depend on data


tancheng:
Current design enables data-arrival-dependent execution. However, the first round/iteration execution shouldn't depend on the arrival of data, in order to make all the tiles be ready to consume coming data/operand/input. We need to follow the same functionality [here](https://github.com/tancheng/VectorCGRA/blob/c07cb57514c3f14e5dc40ae2e56766c76b98a9a5/fu/single/PhiRTL.py#L117) but set the `send_out[0].val` as `1` to yield fake data with predicate as `0`.

tancheng:
<img width="136" alt="Image" src="https://github.com/user-attachments/assets/73141415-e82c-4c40-acb7-8e4b68443848" />

To be more clear, it is not "first round/iteration", it is more like first a few iterations, i.e., prologue cycles. Temporal-scheduled CGRA cannot just depend on data arrival as there is other operations at different cycles (might be used to deliver/produce valid data). The first valid/useful execution of a control signal cannot purely depend on data arrival due to different latency on different operand paths, see `n8` in the `II=3` mapping. Its first valid execution should be on `cycle 10` as it requires `n7` output, i.e., its first 3 times (in the first 9 cycles) executions are invalid.

<img width="89" alt="Image" src="https://github.com/user-attachments/assets/81797dd0-9e52-4e62-af4e-0932855fe632" />

 - Some might think `n8` can just hang there as it didn't receive valid data. This is not okay as some "other operations" mapped on the same tile as `n8` would be blocked. The "other operations" might be the one to produce valid data for `n8`, then `n8` hanging can never be released. This is mainly because the feature of "temporal-scheduling". For "spatial-only-scheduling", there is no such issue, an operation can always depend on data arrival.
 - Therefore, we do need to let `n8` proceed, produce invalid result, and **NOT** consume valid operand/input. In this case, we need to make the `NormalQueue` (buffer of FU) have 3 entries to hold the valid incoming data/operands. (actually, for this case, it is both `n9` and `n9` needs to execute 3 times as prologue, a little bit tricky.)
 - Accordingly, mapping should be aware of the size of buffer, to disallow invalid mapping that the FU buffer overflows.

tancheng:
To serve the "execute `x` times as prologue", I will add `x` in the control signal format... I might choose 3-bit to represent 0-7 for now.

tancheng:
[...]

Issue #70, opened 2025-01-20
[P2] Deliver addr2controller_vector via cpu_pkt


yuqisun:
For `addr2controller_vector` in [controller/ControllerRTL.py](https://github.com/tancheng/VectorCGRA/pull/68/files/9182cfb334be47afbfa8ebebc478ff334b5d9242#diff-2f580b4a845e2ec77388cdddf2e70b2fd8c19e6705933a7f796484f5331bbbd4), should also be delivered via `cpu_pkt`. Refer to [here](https://github.com/tancheng/VectorCGRA/pull/68#discussion_r1921598667).

PR #107, opened 2025-03-30
[feature] Optional execute/route based on prologue count that is pre-…

tancheng:
Allow users to pre-configure the prologue execution for each operation and each routing.
 - Required by temporal mapping. The design now makes the operation execution depend on input data arrival, leading to some tiles blocking the operations that are mapped after the pending operations, detailed in https://github.com/tancheng/VectorCGRA/issues/73.
 - [ ] Test.
 - Also fixed https://github.com/tancheng/VectorCGRA/issues/105.

### DIFF
--- a/cgra/CgraRTL.py
+++ b/cgra/CgraRTL.py
@@ -105,6 +105,10 @@ class CgraRTL(Component):
     s.recv_from_cpu_pkt //= s.controller.recv_from_cpu_pkt
     s.send_to_cpu_pkt //=  s.controller.send_to_cpu_pkt
 
+    # Assigns tile id.
+    for i in range(s.num_tiles):
+      s.tile[i].tile_id //= i
+
     # Connects ring with each control memory.
     for i in range(s.num_tiles):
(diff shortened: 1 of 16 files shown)


rp15:
Could you explain this feature? Is it orthogonal to what we discussed yesterday about making ctrl_count_per_iter = 4, total_ctrl_steps = 4 configurable? I have some changes for that locally. And you are also fixing some of #105, right?

tancheng:
> Could you explain this feature? Is it orthogonal to what we discussed yesterday about making ctrl_count_per_iter = 4, total_ctrl_steps = 4 configurable? I have some changes for that locally. And you are also fixing some of #105, right?

They are orthogonal. And explained in https://github.com/tancheng/VectorCGRA/issues/73. This is probably the last one related to functionality before finalizing Verilog (though not sure whether we still need https://github.com/tancheng/VectorCGRA/issues/70). Hope all following PRs are only about cleanup/bug fixes.

yuqisun (approved):

PR #119, opened 2025-04-12
[feature] Provide new type for cpu->noc xbar

tancheng:
Provide new pkt for controller -> noc xbar, to distinguish the xbar's `dst` with `InterCgraPkt.dst`.
 - [x] We don't need `from_cpu`, instead, leveraging `num_tiles` as `src_tile_id` should be fair enough.
 - [ ] ~~Directly send inter_cgra_pkt to data memory, rather than splitting/slicing it and feeding sub-fields separately.~~
   - Can be done later.
 - [x] Remove unused message types for cleanup.


tancheng:
@yuqisun, you can sync to this PR to try out your multi-cgra systolic array test. It didn't encounter same issue, but didn't finish within 100 cycles.

yuqisun:
> @yuqisun, you can sync to this PR to try out your multi-cgra systolic array test. It didn't encounter same issue, but didn't finish within 100 cycles.

Thanks @tancheng , good now, let me check the _test.

yuqisun (approved):

PR #130, opened 2025-04-16
Translate-Only Util

yo96:
Added a translated-only util function that only performs translation but not import. Import would invoke Verilator to compile the generated Verilog and for large design it would take forever.

### DIFF
--- a/multi_cgra/test/MeshMultiCgraRTL_test.py
+++ b/multi_cgra/test/MeshMultiCgraRTL_test.py
@@ -8,6 +8,7 @@ Author : Cheng Tan
   Date : Jan 8, 2024
 """
 
+from re import sub
 from pymtl3.passes.backends.verilog import (
   VerilogVerilatorImportPass,
   VerilogPlaceholderPass,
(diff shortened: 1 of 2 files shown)


multi_cgra/test/MeshMultiCgraRTL_test.py
```diff
@@ -307,18 +313,25 @@ def test_sim_homo_2x2_2x2(cmdline_opts):
...
-#                       ['UNSIGNED', 'UNOPTFLAT', 'WIDTH', 'WIDTHCONCAT',
-#                        'ALWCOMBORDER'])
-#   th = config_model_with_cmdline_opts(th, cmdline_opts, duts = ['dut'])
-#   th.apply(DefaultPassGroup())
+def _enable_translate_recursively(m):
+  m.set_metadata(VerilogTranslationPass.enable, True)
+  for child in m.get_child_components(repr):
+    _enable_translate_recursively( child )
+
+def translate_model(top, submodules_to_translate):
```
tancheng, line 319: The main purpose here is to skip the `apply(DefaultPassGroup()`, right? And `top.apply(VerilogPlaceholderPass())` is necessary for this purpose? Just curious why we need a placeholderPass here..

yo96: The main purpose is to apply `VerilogTranslationPass` individually, instead of `VerilogTranslationImportPass`.

`VerilogPlaceholderPass` is for embedded Verilog modules, i.e. if you have imported modules in your design `VerilogPlaceholderPass` will pickle its source file I think.

tancheng: Plz change
https://github.com/tancheng/VectorCGRA/blob/c08b879c3942ffaf281b42080b653f7563b3138e/.github/workflows/python-package.yml#L84
to
```
pytest ../multi_cgra/test/MeshMultiCgraRTL_test.py::test_verilog_homo_2x2_4x4 -xvs --test-verilog --dump-vtb --dump-vcd
```
to enable the feature provided in this PR. Thanks!

(now it takes 40m to finish pre-submit, let's see how long it takes once the action is updated.)

tancheng (approved):

PR #151, opened 2025-06-24
Add test harness for parameterizable multi-CGRA functionality for UI.

yuqisun:
Hi,

Let you know this PR is not completed yet, will read properties from multiCgraParam, will merge once all good.

And:
1. This is an entry point called by multi-cgra GUI
2. Will add to action with default param as None
3. Verilog generated and will test synthesis

Thanks,


yuqisun:
> This would only be used for your local branch? And you plan to merge all into master later?
> 
> Plz add description for this PR.
> 
> * It seems this is file is an entry point called by the GUI?
> * Would the test pass when we include it into the Github Action?
> * Would this test's generated Verilog be synthesizable and generating layout?

Hi, added details in description, it's not completed yet, will continue tomorrow.

PR #158, opened 2025-07-16
Issue 55, context switch.

yyan7223:
Updates the logic about how to record and recover the progress (PHI's output i), and how to reset the predicate of FU output.

I tried to write a testbench for the new featrues, but I forgot the tile's internal architecture, I remember there is a picture before that shows the separate crossbar...... 

Can we discuss about this in tomorrow's meeting?

### DIFF
--- a/mem/ctrl/ContextSwitchRTL.py
+++ b/mem/ctrl/ContextSwitchRTL.py
@@ -11,6 +11,7 @@ from pymtl3.stdlib.primitive import *
 from ...lib.basic.val_rdy.ifcs import ValRdyRecvIfcRTL as RecvIfcRTL
 from ...lib.basic.val_rdy.ifcs import ValRdySendIfcRTL as SendIfcRTL
 from ...lib.cmd_type import *
+from ...lib.messages import *
 from ...lib.opt_type import *
 from ...lib.status_type import *
 from ...lib.util.common import *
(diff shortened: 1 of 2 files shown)


yyan7223:
Hi @tancheng ,

Sorry for the late response, I was working on the thesis last week, it is much time consuming than I thought.

Here is a RTL design for `ContextSwitchUnitRTL` you suggested:


The inports include:
- `recv_pkt_from_controller`: Receives `PAUSE` and `RESUME` commands from the controller.
- `recv_opt`: Receives ctrl signals from ctrl mem.
- `ProgIn`: Iteration `i` (progress) from the `PHI_const` operation.

The outports include:
- `ProgOut`: Iteration `i` (progress) to the `PHI_const` operation.
- `ProgOut_valid`: Indicates the output progress is valid, then tile can use it to replace `i` output by `PHI_const`, so as to resume the progress.

The functionalities include:
- Updates the status register `StatReg`: It will be set to `1` after receiving `PAUSE` command, reset to `0` after receiving `RESUME` command, otherwise keeps unchanged.
- Updates the progess register `ProReg`: This register has a default value `-1`, indicating no progress stored. 
    - It will register the `ProgIn` at the rising edge of the incoming clock cycle, only if: (1) current cycle is PHI_const operation `(recv_opt=PHI_const)`; (2) currently under PAUSE status `(status=1)`; (3) ProReg has nothing `ProgOut=-1`.
    - It will be reset to `-1` at the rising edge of the incoming clock cycle, only if: (1) current cycle is PHI_const operation `(recv_opt=PHI_const)`; (2) currently under RESUME status `(status=0)`; (3) ProReg has something`ProgOut != -1`.
        - This condition is also used for setting the signal `Progout_valid=1`, indicating the output progress is valid. Tile can detect this signal and replace the output `i` with the valid `ProgOut`. At next cycle,  `ProgOut` is set to -1 and `Progout_valid` goes back to 0, which means the module is ready for the next `PAUSE` command.
    - It will keep the progress unchanged in other cases.

tancheng:
Hi @yyan7223, thanks for the update~

LGTM.
 - The `ctrl[0]`, `ctrl[1]`, and `ctrl[02]` are not clear to me.
 - Names can be longer and meaningful, e.g., `Prog` -> `progress`, `Stat` -> `status`.
 - `status_register` could hold `idle`, `running`, `pause`, and `resume`. `resume` would only exist for 1 or a few cycles as you described.

yyan7223:
> Hi @yyan7223, thanks for the update~
> 
> LGTM.
> 
> * The `ctrl[0]`, `ctrl[1]`, and `ctrl[02]` are not clear to me.
> * Names can be longer and meaningful, e.g., `Prog` -> `progress`, `Stat` -> `status`.
> * `status_register` could hold `idle`, `running`, `pause`, and `resume`. `resume` would only exist for 1 or a few cycles as you described.

- The `ctrl[0]`, `ctrl[1]`, and `ctrl[02]` are different bits of `Ctrl` that controls two `Mux` in the figure.
- Yeah sure.
- Currently the `status_register` is only 1 bit, `pause->1` and `resume->0`. Seems these two status are enough for save and recovering the progress?

tancheng:
> Currently the status_register is only 1 bit, pause->1 and resume->0. Seems these two status are enough for save and recovering the progress?

Let's make it meaningful. `cmd PAUSE` leads to `PAUSING` status, `cmd RESUME` leads to `RESUMING` status, after `RESUME`, it changes to `STABLE` or whatever. These `ING`s indicate the status of the switch unit.

Issue #149, opened 2025-06-18
[P1] Introduce grant operations


tancheng:
Based on the IR representation (https://github.com/coredac/dataflow/issues/45), I feel grant operations are necessary to serve as first-class basic operations that supported in our HW.
 - `grant_predicate`
 - `grant_once`
 - `grant_always`

With these, we might not need to deliver additional predicate operand towards `phi`, as the predicate bit embedded inside the original operand value could be enough. Actually I need to think about this further. @yuqisun do you remember we discussed the predicated execution for the if/else paths? Do you still have the diagram? I remember you drew sth when we discussed this.

tancheng:
Right, let me think about it :-)

tancheng:
hmm, I don't think we [receive_predicate](https://github.com/tancheng/VectorCGRA/blob/bddd1d500ddd5d93d3c65710161a95979980c180/fu/single/AdderRTL.py#L87) is necessary anymore, we can deprecate it once the `grant_xxx` operation is supported.

Issue #27, opened 2024-11-23
[P1] Predicate should be associated with specific control signal


tancheng:
Existing predicate is limited as only one predicate is allowed in one tile across all cycles (mapper might not have this contraint IIRC), which is not scalable for control-flow (and nested-loop) rich kernels. A simple LUT can make multiple-predicate possible.

PR #164, opened 2025-07-20
Deprecate predicate channel and predicate bit in opcode

tancheng:
We now already support `grant` operation in https://github.com/tancheng/VectorCGRA/issues/149, so predicate can be purely embedded/represented in data. We don't need explicit predicate channel and predicate bit inside the opcode: https://github.com/tancheng/VectorCGRA/issues/27.


tancheng:
Hi @yuqisun, can you help check what is the issue causing the error in the `Github Action` for translating `TileRTL_test.py`: https://github.com/tancheng/VectorCGRA/blob/a32783d338c280db0ad89f6b6c6cc3f95c211a35/.github/workflows/python-package.yml#L68?
 - Reproduce on your side with `--tb=short`
 - and check whether there are other issues?

Sth wrong with my local environment, plz help, thanks~!

yuqisun (approved):

PR #163, opened 2025-07-20
Provide more prologue counters for ctrl_mem and crossbars

tancheng:
More prologue counters enable flexible control of FUs, and routing/FU xbars.
 - Before this PR, one tile has only one counter, which lacks support for different operations in time domain

This PR can close https://github.com/tancheng/VectorCGRA/issues/27 and https://github.com/tancheng/VectorCGRA/issues/156

TODO in this PR:
 - ~~Refactor `phi_const` to avoid retrieving data from the input channel: https://github.com/tancheng/VectorCGRA/blob/a32783d338c280db0ad89f6b6c6cc3f95c211a35/fu/single/PhiRTL.py#L110~~
   - We are not able to do this, as the `phi_const` or `first` cannot/shouldn't be seen by the `routing_xbar`

PR #167, opened 2025-08-03
Decrease mem size to speedup synthesize

yuqisun:
image


yuqisun:
The long-time running issue is fixed by https://github.com/tancheng/VectorCGRA/pull/163

PR #171, opened 2025-08-12
Implementation of context switch module and its testcase.

yyan7223:
Here is the RTL design for `ContextSwitchUnitRTL` (blue) and how it is connected with other components within the tile:

The inports include:
- `recv_pkt_from_controller`: Receives `PAUSE` and `RESUME` commands from the controller.
- `recv_opt`: Receives ctrl signals from ctrl mem.
- `progess_in`: `iteration` (progress) from the `PHI_const` operation executed in `functional unit (FU)`.

The outports include:
- `progess_out`: `iteration` that replaces the `PHI_const` operation of `FU`, so as to resume the progress.
- `progess_out_vld`: Indicates when the replacment occurs.

The functionalities include:
- Updates the status register `statusReg`: It will be set to `PAUSING=2'b11` after receiving `PAUSE` command, reset to `RESUMING=2'b10` after receiving `RESUME` command, otherwise in `RUNNING=2'b00`.
- Updates the progess register `progressReg`: This register has a default value `0`, indicating no progress stored. 
    - It will register the `progess_in` at the rising edge of the incoming clock cycle, only if: (1) current cycle is PHI_const operation `(recv_opt=PHI_const)`; (2) currently under PAUSING status `(statusReg=2'b11)`; (3) progressReg has nothing `progess_out=0`.
    - It will be reset to `0` at the rising edge of the incoming clock cycle, only if: (1) current cycle is PHI_const operation `(recv_opt=PHI_const)`; (2) currently under RESUMING status `(statusReg=2'b10)`; (3) progressReg has something`progess_out != 0`.
        - This condition is also used for setting the signal `progess_out.predicate=1`, indicating the output progress is true. Tile can detect the predicate bit and replace the output `iteration` of `FU` with the valid `progess_out`. At next cycle,  `progess_out` is set back to 0, which means the module is ready for the next `PAUSE` command.
    - It will keep the progress unchanged in other cases.

### DIFF
--- a/mem/ctrl/ContextSwitchRTL.py
+++ b/mem/ctrl/ContextSwitchRTL.py
@@ -22,7 +22,7 @@ class ContextSwitchRTL(Component):
     # Constant
     CmdType = mk_bits(clog2(NUM_CMDS))
     StatusType = mk_bits(clog2(NUM_STATUS))
-    DataType = mk_bits(clog2(data_nbits))
+    DataType = mk_bits(data_nbits)
     OptType = mk_bits(clog2(NUM_OPTS))
 
     # Interface
@@ -51,11 +51,12 @@ class ContextSwitchRTL(Component):
       s.is_executing_phi @= (s.recv_opt == OPT_PHI_CONST)
 
       # Updates progress_out with the recorded progress.
-      s.progress_out @= s.progress_reg
       if (~s.progress_is_null & s.is_resuming & s.is_executing_phi):
         s.progress_out_vld @= 1
+        s.progress_out @= s.progress_reg
       else:
         s.progress_out_vld @= 0
+        s.progress_out @= DataType(0)
 
     @update_ff
     def update_regs():
--- a/mem/ctrl/ContextSwitchRTL.py
+++ b/mem/ctrl/ContextSwitchRTL.py
@@ -28,6 +28,8 @@ class ContextSwitchRTL(Component):
     # Interface
     s.recv_cmd = InPort(CmdType)
     s.recv_cmd_vld = InPort(b1)
+    # We don't need recv_opt_vld as progress_in_vld is enough 
+    # to tell whether the opt is valid.
     s.recv_opt = InPort(OptType)
     s.progress_in = InPort(DataType)
     s.progress_in_vld = InPort(b1)
--- a/mem/ctrl/ContextSwitchRTL.py
+++ b/mem/ctrl/ContextSwitchRTL.py
@@ -1,0 +1,95 @@
+"""
+==========================================================================
+ContextSwitchRTL.py
+==========================================================================
+Records/resumes progress (itertion) for functional unit (FU).
+
+Author : Yufei Yang
+  Date : Aug 11, 2025
+"""
+from pymtl3.stdlib.primitive import *
+from ...lib.basic.val_rdy.ifcs import ValRdyRecvIfcRTL as RecvIfcRTL
+from ...lib.basic.val_rdy.ifcs import ValRdySendIfcRTL as SendIfcRTL
+from ...lib.cmd_type import *
+from ...lib.opt_type import *
+from ...lib.util.common import *
+
+class ContextSwitchRTL(Component):
+
+  def construct(s, CgraPayloadType, DataType, CtrlType, num_status = 3):
+
+    # Constant
+    StatusType = mk_bits(clog2(num_status))
+
+    # Interface
+    s.recv_cmd = RecvIfcRTL(CgraPayloadType)
+    s.recv_opt = RecvIfcRTL(CtrlType)
+    s.progress_in = RecvIfcRTL(DataType)
+    s.progress_out = SendIfcRTL(DataType)
+   
+    # Component
+    s.progress_reg = Reg(DataType)
+    s.status_reg = Reg(StatusType)
+    #s.condition = Wire(Bits3)
+    # How to create a wire that can be bit sliced?
+    s.condition_2 = Wire(b1)
+    s.condition_1 = Wire(b1)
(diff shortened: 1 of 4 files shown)


mem/ctrl/test/ContextSwitchRTL_test.py
```diff
@@ -0,0 +1,162 @@
...
+                     num_tile_inports,
+                     num_tile_outports,
+                     num_registers_per_reg_bank)
+
+  CgraPayloadType = mk_cgra_payload(DataType,
+                                    DataAddrType,
+                                    CtrlType,
+                                    CtrlAddrType)
+
+  src_cmds = [# Simulates the PAUSE command to record the progress.
```
tancheng, line 93: Plz add some comments above to explain (step by step) what we are testing here.

mem/ctrl/ContextSwitchRTL.py
```diff
@@ -0,0 +1,95 @@
...
+  def construct(s, CgraPayloadType, DataType, CtrlType, num_status = 3):
+
+    # Constant
+    StatusType = mk_bits(clog2(num_status))
+
+    # Interface
+    s.recv_cmd = RecvIfcRTL(CgraPayloadType)
+    s.recv_opt = RecvIfcRTL(CtrlType)
+    s.progress_in = RecvIfcRTL(DataType)
+    s.progress_out = SendIfcRTL(DataType)
```
tancheng, line 28: For these three:
```
    s.recv_opt = RecvIfcRTL(CtrlType)
    s.progress_in = RecvIfcRTL(DataType)
    s.progress_out = SendIfcRTL(DataType)
```
Should we just use `InPort` or `RecvIfc`? I think `InPort` might be easier so we don't need to check val/rdy?

mem/ctrl/ContextSwitchRTL.py
```diff
@@ -0,0 +1,95 @@
...
+        s.status_reg.in_ @= StatusType(0)
+      else:
+        # Keeps unchanged.
+        s.status_reg.in_ @= s.status_reg.out
+
+      # Updates the progress register.
+      s.progress_in.rdy @= 1
+      if (s.condition_2 & s.condition_1 & s.condition_0 & s.progress_in.val):
+        # Records the progress.
+        s.progress_reg.in_ @= s.progress_in.msg
```
tancheng, line 72: I believe `progress_in` will be connected to: https://github.com/tancheng/VectorCGRA/blob/96b8adfa46a51b74c6913497e141b667d5c4bf08/mem/ctrl/CtrlMemDynamicRTL.py#L55

 - Am I right? I am confused why you declare it with `DataType` though.
 - And do you think only `progress` itself is enough to resume a paused task, do we need all the other status in the ctrl mem? (open question, no need to complete in current PR): https://github.com/tancheng/VectorCGRA/blob/96b8adfa46a51b74c6913497e141b667d5c4bf08/mem/ctrl/CtrlMemDynamicRTL.py#L62-L66

mem/ctrl/ContextSwitchRTL.py
```diff
@@ -0,0 +1,95 @@
...
+      # Updates recv_opt.
+      s.recv_opt.rdy @= 1
+
+      # Updates recv_cmd
+      s.recv_cmd.rdy @= 1
+
+      # Updates the status register.
+      if (s.recv_cmd.val & (s.recv_cmd.msg.cmd == CMD_PAUSE)):
+        # PAUSING: 2'b11
+        s.status_reg.in_ @= StatusType(3)
```
tancheng, line 57: this `s.status_reg` is updated like a combinational logic, rather than "on_edge", I can't remember it clearly, but you mean as long as we set its `.in_`, the content would be updated/available in next cycle?

Why not just declare them as `Wire`, and update them in an `@update_ff`? They would be totally under your control, and will be synthesized as registers.

mem/ctrl/ContextSwitchRTL.py
tancheng: I can't recall that, do we need both iterator and the phi_const's value? It is not clear in your PR's description, can we try to sync again on this? It has been too long. Plz provide ETA, if you are not able to make it, plz let me know.

mem/ctrl/ContextSwitchRTL.py
```diff
@@ -0,0 +1,95 @@
...
+      elif (s.recv_cmd.val & (s.recv_cmd.msg.cmd == CMD_RESUME)):
+        # RESUMING: 2'b10
+        s.status_reg.in_ @= StatusType(2)
+      elif (s.recv_cmd.val & (s.recv_cmd.msg.cmd == CMD_LAUNCH)):
+        # RUNNING: 2'b00
+        s.status_reg.in_ @= StatusType(0)
```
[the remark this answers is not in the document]
yyan7223: PAUSING RESUMING RUNNING IDLE

mem/ctrl/ContextSwitchRTL.py
```diff
@@ -0,0 +1,91 @@
...
+    StatusType = mk_bits(clog2(NUM_STATUS))
+    DataType = mk_bits(data_nbits)
+    OptType = mk_bits(clog2(NUM_OPTS))
+
+    # Interface
+    s.recv_cmd = InPort(CmdType)
+    s.recv_cmd_vld = InPort(b1)
+    s.recv_opt = InPort(OptType)
+    s.progress_in = InPort(DataType)
+    s.progress_in_vld = InPort(b1)
```
tancheng, line 33: Add a comment above mentioning "We don't need recv_opt_vld as progress_in_vld is enough to tell whether the opt is valid"?

And this PR is ready for review?

tancheng (approved):

Issue #55, opened 2025-01-03
[P1] Support context switch for ctrl memory


tancheng:

tancheng:
Ctrl Memory by default iterates from `0` to `num_of_items - 1`, and repeats that again and again.
We can provide prologue index and epilogue index, so the ctrl memory can point `0` to `prologue_index` once, repeat `prologue_index` to `epilogue_index` until `return` is encountered, and then perform `epilogue_index` to `num_of_items - 1` once.

yyan7223:
Sure, but I'm curious about what is that used for?

tancheng:
This issue is mainly to support multi-kernel/multi-context switch (via `CMD`).

tancheng:
Instead of `prologue` and `epilogue`, we can use `CMD` with `kernel_id` to switch between different kernels.

tancheng:
instead of prologue/epilogue, we can support dynamic context switch between different sets of control signals.

tancheng:
Hi @yyan7223, I believe supporting context switching across kernels could also be a baseline for your work.

tancheng:
Hi @yyan7223, what is your current priority among issues?

yyan7223:
> Hi [@yyan7223](https://github.com/yyan7223), what is your current priority among issues?

I haven't considered about this, do I need to work on this issue first, or I start with issue #26 and #74 that have P1?

yyan7223:
Do we still have this issue for fuse execution mode? In my design, the ctrl memory address will always increase by 1 / clock cycle no matter whether `ld` miss or hit. So it seems now we don't need context switch, or it services for the new functionality?

tancheng:
> Do we still have this issue for fuse execution mode? In my design, the ctrl memory address will always increase by 1 / clock cycle no matter whether `ld` miss or hit. So it seems now we don't need context switch, or it services for the new functionality?

New functionality. I was thinking user or certain engine can issue "context switch" command. But anyways, you are right, I think https://github.com/tancheng/VectorCGRA/issues/74 should have higher priority for you, as it is a fundamental part of your paper/idea.

tancheng:
TODO @yyan7223:
 - [ ] PR1: For now, ctrl memory iterating 0 -> bound, bound indicates how many ctrl signals are stored.
[...]

PR #172, opened 2025-08-23
Issue #55: Supplement the output predicate for contextswitch module.

yyan7223:
The output predicate should be performed AND operation with the predicate of FU's output, so as to get the real predicate.
Sets predicate=0 only when in the PAUSING status and everytime executing the PHI_CONST (first node in DFG). Then predicate=0 will be broadcasted to all other operations in this iteration via the dataflow. Consequently, all iterations after progress is recorded all have predicate=0.

### DIFF
--- a/mem/ctrl/ContextSwitchRTL.py
+++ b/mem/ctrl/ContextSwitchRTL.py
@@ -35,6 +35,9 @@ class ContextSwitchRTL(Component):
     s.progress_in_vld = InPort(b1)
     s.progress_out = OutPort(DataType)
     s.progress_out_vld = OutPort(b1)
+    # The output predicate should be performed AND operation
+    # with the predicate of FU's output, so as to get the real predicate.
+    s.predicate = OutPort(b1)
    
     # Component
     s.progress_reg = Wire(DataType)
@@ -59,6 +62,16 @@ class ContextSwitchRTL(Component):
       else:
         s.progress_out_vld @= 0
         s.progress_out @= DataType(0)
+
+      # Sets predicate=0 only when in the PAUSING status and
+      # everytime executing the PHI_CONST (first node in DFG). 
+      # Then predicate=0 will be broadcasted to all other operations
+      # in this iteration via the dataflow. Consequently, all 
+      # iterations after progress is recorded all have predicate=0.
+      if (s.is_pausing & s.is_executing_phi):
+        s.predicate @= 0
+      else:
+        s.predicate @= 1
 
     @update_ff
     def update_regs():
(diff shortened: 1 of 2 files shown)


mem/ctrl/ContextSwitchRTL.py
```diff
@@ -22,22 +23,21 @@ def construct(s, data_nbits):
...
 
     # Interface
     s.recv_cmd = InPort(CmdType)
     s.recv_cmd_vld = InPort(b1)
-    # We don't need recv_opt_vld as progress_in_vld is enough 
+    # We don't need recv_opt.predicate as progress_in.predicate is enough 
     # to tell whether the opt is valid.
     s.recv_opt = InPort(OptType)
     s.progress_in = InPort(DataType)
-    s.progress_in_vld = InPort(b1)
```
tancheng, line 35: Why remove this? `progress_in.predicate` is different from `valid`, the output from the FU could be `valid` but the predicate could be 0.

mem/ctrl/ContextSwitchRTL.py
```diff
@@ -50,28 +50,24 @@ def construct(s, data_nbits):
...
-      # iterations after progress is recorded all have predicate=0.
+      # The output of PHI_CONST (first node in DFG) during the PAUSING
+      # status should always have predicate=0, as it will be broadcasted 
+      # to all other operations in this iteration via the dataflow. 
       if (s.is_pausing & s.is_executing_phi):
-        s.predicate @= 0
+        s.reset_fu_output_predicate @= 1
       else:
-        s.predicate @= 1
+        s.reset_fu_output_predicate @= 0
```
tancheng, line 70: Should these be:
```
      if (s.is_pausing & s.is_executing_phi):
        s.progress_out.predicate @= 1
      else:
        s.progress_out.predicate @= 0
```
Why you want to keep a `reset_fu_output_predicate`? And I think `progress_out_vld` should exist. Again, `predicate` is not `valid`, `predicate` is for dataflow's execution to handle ctrl-flow; `valid` is the HW concept in terms of handshake.

mem/ctrl/ContextSwitchRTL.py
yyan7223: Emm...If the output progress from FU has predicate=0, it is meaningless to contextswitch module.

mem/ctrl/ContextSwitchRTL.py
yyan7223: Hi cheng, 

Tile can use signal `reset_fu_output_predicate` to set `predicate=0` for incoming iterations. I've commented the meaning of this signal in this commit.
And yes, `predicate` is not `valid`, but does the context switch module requires handshake? What if it only replaces FU's output.msg without changing its `val` or `rdy`?

Regards,
Yufei

tancheng: hmm,
> How about we just use this module to replace FU's output.msg without changing its val or rdy?

That's a good idea. let me look into the code again.

PR #182, opened 2025-09-08
Clears the channel by resetting counts.

yyan7223:
Hi @tancheng ,

May I have the request to commit to PyOCN? 
I add a count_reset inport to `ChannelRTL` in PyOCN to support channel clearing.

Regards,
Yufei


tancheng:
Can we have a different `ChannelResetRTL.py` duplicated from pyocn folder for this feature?

PR #208, opened 2025-10-25
FIR kernel, run->pause->resume.

yyan7223:
We can schedule a meeting to check the timing of pausing and resuming behavior.

### DIFF
--- a/cgra/test/CgraWithContextSwitchRTL_test.py
+++ b/cgra/test/CgraWithContextSwitchRTL_test.py
@@ -486,9 +486,6 @@ def sim_fir_return(cmdline_opts, mem_access_is_combinational):
                            payload = CgraPayloadType(CMD_CONFIG_PROLOGUE_FU, ctrl_addr = 0,
                                                      data = DataType(1, 1))),
           IntraCgraPktType(0, 0,
-                           payload = CgraPayloadType(CMD_CONFIG_PROLOGUE_FU, ctrl_addr = 1,
-                                                     data = DataType(1, 1))),
-          IntraCgraPktType(0, 0,
                            payload = CgraPayloadType(CMD_CONFIG_PROLOGUE_ROUTING_CROSSBAR, ctrl_addr = 0,
                                                      ctrl = CtrlType(routing_xbar_outport = [
                                                         TileInType(0), TileInType(0), TileInType(0), TileInType(0),
@@ -496,12 +493,6 @@ def sim_fir_return(cmdline_opts, mem_access_is_combinational):
                                                      data = DataType(1, 1))),
           IntraCgraPktType(0, 0,
                            payload = CgraPayloadType(CMD_CONFIG_PROLOGUE_FU_CROSSBAR, ctrl_addr = 0,
-                                                     ctrl = CtrlType(fu_xbar_outport = [
-                                                        FuOutType(0), FuOutType(0), FuOutType(0), FuOutType(0),
-                                                        FuOutType(0), FuOutType(0), FuOutType(0), FuOutType(0)]),
-                                                     data = DataType(1, 1))),
-          IntraCgraPktType(0, 0,
-                           payload = CgraPayloadType(CMD_CONFIG_PROLOGUE_FU_CROSSBAR, ctrl_addr = 1,
                                                      ctrl = CtrlType(fu_xbar_outport = [
                                                         FuOutType(0), FuOutType(0), FuOutType(0), FuOutType(0),
                                                         FuOutType(0), FuOutType(0), FuOutType(0), FuOutType(0)]),
@@ -695,7 +686,7 @@ def sim_fir_return(cmdline_opts, mem_access_is_combinational):
                                                                        FuOutType(1), FuOutType(0), FuOutType(0), FuOutType(0)],
                                                                       write_reg_from = write_reg_from_code))),
 
-          # NAH.
+          # NOT.
           IntraCgraPktType(0, 5,
                             payload = CgraPayloadType(CMD_CONFIG, ctrl_addr = 3,
                                                       ctrl = CtrlType(OPT_NOT,
@@ -887,13 +878,12 @@ def sim_fir_return(cmdline_opts, mem_access_is_combinational):
 
       # Resumes tile 0's execution.
       [ 
-        # Resends the const for PHI_COSNT with predicate = 0 to correctly resume the progress.
-#        IntraCgraPktType(0, 0, payload = CgraPayloadType(CMD_CONST, data = DataType(kSumInitValue, 0))),
+        # Resends const for PHI_COSNT.
+        IntraCgraPktType(0, 0, payload = CgraPayloadType(CMD_CONST, data = DataType(kSumInitValue, 1))),
         # Resets ctrl mem raddr.
         IntraCgraPktType(0, 0, payload = CgraPayloadType(CMD_CONFIG_CTRL_LOWER_BOUND, data = DataType(0, 1))),
         # Resets FU's prologue
         IntraCgraPktType(0, 0, payload = CgraPayloadType(CMD_CONFIG_PROLOGUE_FU, ctrl_addr = 0, data = DataType(1, 1))),
-        IntraCgraPktType(0, 0, payload = CgraPayloadType(CMD_CONFIG_PROLOGUE_FU, ctrl_addr = 1, data = DataType(1, 1))),
         # Resumes execution.
         IntraCgraPktType(0, 0, payload = CgraPayloadType(CMD_RESUME))
       ],
@@ -912,7 +902,7 @@ def sim_fir_return(cmdline_opts, mem_access_is_combinational):
       # Restarts tile 4's execution.
       [
         # Resends const for ADD_CONST.
- #       IntraCgraPktType(0, 4, payload = CgraPayloadType(CMD_CONST, data = DataType(kCoefficientBaseAddress, 1))),
+        IntraCgraPktType(0, 4, payload = CgraPayloadType(CMD_CONST, data = DataType(kCoefficientBaseAddress, 1))),
         # Resets ctrl mem raddr.
         IntraCgraPktType(0, 4, payload = CgraPayloadType(CMD_CONFIG_CTRL_LOWER_BOUND, data = DataType(0, 1))),
         # Restarts execution.
@@ -922,7 +912,7 @@ def sim_fir_return(cmdline_opts, mem_access_is_combinational):
       # Restarts tile 5's execution.
       [
         # Resends const for CMP.
-  #      IntraCgraPktType(0, 5, payload = CgraPayloadType(CMD_CONST, data = DataType(kLoopUpperBound, 1))),
+        IntraCgraPktType(0, 5, payload = CgraPayloadType(CMD_CONST, data = DataType(kLoopUpperBound, 1))),
         # Resets ctrl mem raddr.
         IntraCgraPktType(0, 5, payload = CgraPayloadType(CMD_CONFIG_CTRL_LOWER_BOUND, data = DataType(0, 1))),
         # Restarts execution.
@@ -931,10 +921,10 @@ def sim_fir_return(cmdline_opts, mem_access_is_combinational):
 
       # Resumes tile 8's execution.
       [
-        # Resends the const for PHI_COSNT with predicate = 0 to correctly resume the progress.
-   #     IntraCgraPktType(0, 8, payload = CgraPayloadType(CMD_CONST, data = DataType(kLoopLowerBound, 0))),
+        # Resends const for PHI_COSNT.
+        IntraCgraPktType(0, 8, payload = CgraPayloadType(CMD_CONST, data = DataType(kLoopLowerBound, 1))),
         # Resends const for ADD_CONST.
-   #     IntraCgraPktType(0, 8, payload = CgraPayloadType(CMD_CONST, data = DataType(kInputBaseAddress, 1))),
+        IntraCgraPktType(0, 8, payload = CgraPayloadType(CMD_CONST, data = DataType(kInputBaseAddress, 1))),
         # Resets ctrl mem raddr.
         IntraCgraPktType(0, 8, payload = CgraPayloadType(CMD_CONFIG_CTRL_LOWER_BOUND, data = DataType(0, 1))),
         # Resumes execution.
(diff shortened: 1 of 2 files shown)


fu/single/SelRTL.py
```diff
@@ -51,6 +51,9 @@ def construct(s, DataType, PredicateType, CtrlType,
     s.to_mem_waddr = SendIfcRTL(AddrType)
     s.to_mem_wdata = SendIfcRTL(DataType)
 
+    # Redundant interface for PhiRTL
+    s.clear = InPort(b1)
```
tancheng: Comment could be `Redundant interface, only used by PhiRTL`.

And sorry about that but this might be added to all FU..
 - All basic `XxxFuRTL` inherited from `FuRTL`, e.g., [AdderRTL](https://github.com/tancheng/VectorCGRA/blob/f826a1dbdb86cefcfdb6bcf8c891af52560fa6f3/fu/single/AdderRTL.py#L15), [CmpRTL](https://github.com/tancheng/VectorCGRA/blob/f826a1dbdb86cefcfdb6bcf8c891af52560fa6f3/fu/single/CompRTL.py#L15), this `s.clear` can be added inside the [FuRTL](https://github.com/tancheng/VectorCGRA/blob/master/fu/basic/Fu.py).
 - Some others like this [SelRTL](https://github.com/tancheng/VectorCGRA/blob/f826a1dbdb86cefcfdb6bcf8c891af52560fa6f3/fu/single/SelRTL.py#L17) that is not inherited from FuRTL, needs to add this `s.clear` as you do now.
 - For [FpAddRTL](https://github.com/tancheng/VectorCGRA/blob/master/fu/float/FpAddRTL.py) and [FusedFuRTL](https://github.com/tancheng/VectorCGRA/blob/master/fu/basic/TwoPrlCombo.py), this `s.clear` is also needed. Though is won't be used (i.e., "redundant", but need to make the interfaces be same, so they can "plug-in"...)

mem/ctrl/CtrlMemDynamicRTL.py
```diff
@@ -200,11 +200,8 @@ def update_whether_we_can_iterate_ctrl():
             s.start_iterate_ctrl <<= 1
     # TODO: issue #191, stop iterate ctrl after 10 cycels during pausing status, 
     # so as to clear channels safely.
-          # Deleted by yufeiyang on Sep 25, 2025, as it has conflict with ContextSwitchRTL.
-          #elif s.recv_pkt_queue.send.msg.payload.cmd == CMD_TERMINATE:
-          #  s.start_iterate_ctrl <<= 0
-          #elif s.recv_pkt_queue.send.msg.payload.cmd == CMD_PAUSE:
-          #  s.start_iterate_ctrl <<= 0
+          elif s.recv_pkt_queue.send.msg.payload.cmd == CMD_TERMINATE:
```
tancheng: No need `s.start_iterate_ctrl <<= 0` for `clear`?

mem/ctrl/CtrlMemDynamicRTL.py
```diff
@@ -228,6 +225,8 @@ def update_raddr_and_fu_prologue():
           s.prologue_count_reg_fu[i] <<= 0
       elif s.recv_pkt_queue.send.val & (s.recv_pkt_queue.send.msg.payload.cmd == CMD_CONFIG_CTRL_LOWER_BOUND):
         s.reg_file.raddr[0] <<= trunc(s.recv_pkt_queue.send.msg.payload.data.payload, CtrlAddrType)
+      elif s.recv_pkt_queue.send.val & (s.recv_pkt_queue.send.msg.payload.cmd == CMD_CLEAR):
+        s.times <<= TimeType(0)
```
tancheng: No need `s.times <<= TimeType(0)` for `terminate`?

fu/single/SelRTL.py
tancheng: And `clear` also needs to clear this register: https://github.com/tancheng/VectorCGRA/blob/f826a1dbdb86cefcfdb6bcf8c891af52560fa6f3/fu/single/GrantRTL.py#L129

cgra/test/CgraRTL_fir_test_temporal.py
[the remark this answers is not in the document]
yyan7223: Yeah it's reduntant, have removed it in the latest commit.

cgra/test/CgraWithContextSwitchRTL_test.py
```diff
@@ -0,0 +1,990 @@
...
+                                                                      FuOutType(0), FuOutType(0), FuOutType(0), FuOutType(0)],
+                                                                     read_reg_from = [b1(0), b1(1), b1(0), b1(0)]))),
+
+          # Launch the tile.
+          IntraCgraPktType(0, 9, payload = CgraPayloadType(CMD_LAUNCH))
+      ],
+
+      # Let all kernels free running for 6 cycles (too long will block the const mem of tile 15).
+      # We send useless data to unused tile 15 to simulate the free-running.
+      [ IntraCgraPktType(0, 15, payload = CgraPayloadType(CMD_CONST, data = DataType(0, 0))) for _ in range(12) ],
```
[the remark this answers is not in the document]
yyan7223: Have fixed the comments

cgra/test/CgraWithContextSwitchRTL_test.py
```diff
@@ -0,0 +1,990 @@
...
+        IntraCgraPktType(0, 9, payload = CgraPayloadType(CMD_CLEAR)),
+        # Restarts execution.
+        IntraCgraPktType(0, 9, payload = CgraPayloadType(CMD_LAUNCH))
+      ]
+  ]
+
+  
+  src_query_pkt = \
+      [
+          # IntraCgraPktType(payload = CgraPayloadType(CMD_LOAD_REQUEST, data_addr = kStoreAddress)),
```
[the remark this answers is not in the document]
yyan7223: This is originally commented in `sim_fir_return`. It uses `expected_complete_sink_out_pkg` to valid the correctness.

mem/ctrl/CtrlMemDynamicRTL.py
yyan7223: Maybe we should not iterate control signals after receving `CMD_TERMINATE`?

mem/ctrl/CtrlMemDynamicRTL.py
yyan7223: Maybe we should clear something after receving `CMD_CLEAR`?

tancheng: I didn't get this. Now the `CMD_CLEAR` and `CMD_TERMINATE` are the similar thing..

mem/ctrl/test/ContextSwitchRTL_test.py
```diff
@@ -258,7 +260,28 @@ def test_preserve_resume_accumulation():
               CmdType(CMD_CONST)  # cycle 13
               ]
   
-  # All inputs provided by CtrlMemDynamicRTL have 1-cycle delay because of recv_pkt_queue.
+  # Preloads the address 2 of the PHI_CONST.
+  src_phi_addr = [
+                         # Starts execution.
```
tancheng: ditto

mem/ctrl/CtrlMemDynamicRTL.py
```diff
@@ -135,11 +135,12 @@ def update_msg():
          (s.recv_pkt_queue.send.msg.payload.cmd == CMD_LAUNCH) | \
          (s.recv_pkt_queue.send.msg.payload.cmd == CMD_TERMINATE) | \
          (s.recv_pkt_queue.send.msg.payload.cmd == CMD_PAUSE) | \
+         (s.recv_pkt_queue.send.msg.payload.cmd == CMD_PRESERVE) | \
```
tancheng: Can we rename `CMD_PRESERVE` to `CMD_PRESERVE_PHI`?

fu/single/MemUnitRTL.py
```diff
@@ -114,13 +117,21 @@ def comb_logic():
...
-          s.to_mem_raddr.val @= s.recv_all_val & ~s.already_sent_raddr
+          # Do not access memory if the raddr has predicate=0.
+          if s.recv_in[s.in0_idx].msg.predicate == 0:
+            s.to_mem_raddr.val @= 0
+          else:
+            s.to_mem_raddr.val @= s.recv_all_val & ~s.already_sent_raddr
           s.from_mem_rdata.rdy @= s.send_out[0].rdy
           # FIXME: As the memory access might take more than one cycle,
           # the send_out valid no need to depend on recv_all_val.
           s.send_out[0].val @= s.from_mem_rdata.val
```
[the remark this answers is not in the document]
yyan7223: Have fixed this issue in the latest commit.

In previous code, memory does not return a fake data with `predicate=0` when raddr has a `predicate=0`, then the consumer `X` will block tile 4 due to the lack of `LD` results. However, this has no effect to the pausing behaviour because tile 4 does not mapped with operations that belong to initiated iterations, so that's why it worked. 

When will the previous code do not work? If tile 4 blocks with `X` of iteration 4, but the input channel has `LD` results of iteration 3 have not been consumed, then pausing will fail no matter how long we wait. But does this scenario really exist?

But anyway, I have modified the code to what we originally designed, i.e., memory returns a fake data with `predicate=0` when raddr has a `predicate=0`. It does not work initially because there is a deadlock when assigning values to `Fu`, `fu_crossbar`, and `routing_crossbar`:
(1) Setting `recv_opt.val` relies on `xxx_done`;

(2) Setting `xxx_done` relies on `recv_opt.rdy`;

(3) But setting `recv_opt.rdy` relies on `recv_opt.val`, so there is a deadlock.

WDYT?

I will fix other issues tomorrow.

tile/TileWithContextSwitchRTL.py
```diff
@@ -308,9 +308,12 @@ def update_opt():
 
       # FIXME: Do we still need separate element and routing_xbar?
       # FIXME: Do we need to consider reg bank here?
-      s.element.recv_opt.val @= s.ctrl_mem.send_ctrl.val & ~s.element_done
-      s.routing_crossbar.recv_opt.val @= s.ctrl_mem.send_ctrl.val & ~s.routing_crossbar_done
-      s.fu_crossbar.recv_opt.val @= s.ctrl_mem.send_ctrl.val & ~s.fu_crossbar_done
+      #s.element.recv_opt.val @= s.ctrl_mem.send_ctrl.val & ~s.element_done
+      #s.routing_crossbar.recv_opt.val @= s.ctrl_mem.send_ctrl.val & ~s.routing_crossbar_done
+      #s.fu_crossbar.recv_opt.val @= s.ctrl_mem.send_ctrl.val & ~s.fu_crossbar_done
```
tancheng: We need to keep these as they are.

fu/single/MemUnitRTL.py
```diff
@@ -117,20 +117,24 @@ def comb_logic():
...
-          # the send_out valid no need to depend on recv_all_val.
-          s.send_out[0].val @= s.from_mem_rdata.val
-          s.send_out[0].msg @= s.from_mem_rdata.msg
-          # If raddr has predicate=0, the predicate of rdata should also be 0.
+          # Although we do not access memory when raddr has predicate=0,
+          # we still need to simulate that memory returns a fake data with predicate=0,
+          # so that the consumer will not block due to the lack of data.
+          # Then all initiated iterations can be normally drained.
           if s.recv_in[s.in0_idx].msg.predicate == 0:
+            s.send_out[0].val @= 1
```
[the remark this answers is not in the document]
yyan7223: Have changed in the latest commit.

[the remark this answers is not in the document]
tancheng: I didn't see any changes happened. `s.send_out[0].val @= s.recv_all_val`

tile/TileWithContextSwitchRTL.py
tancheng: same here

lib/cmd_type.py
```diff
@@ -64,7 +63,6 @@
   CMD_GLOBAL_REDUCE_MUL:                "(GLOBAL_REDUCE_MUL)",
   CMD_GLOBAL_REDUCE_ADD_RESPONSE:       "(GLOBAL_REDUCE_ADD_RESPONSE)",
   CMD_GLOBAL_REDUCE_MUL_RESPONSE:       "(GLOBAL_REDUCE_MUL_RESPONSE)",
-  CMD_PRESERVE:                         "(PRESERVE_SUM_VALUE)",
-  CMD_CLEAR:                            "(CLEAR_CHANNELS)"
+  CMD_PRESERVE:                         "(PRESERVE_SUM_VALUE)"
```
tancheng: `PRESERVE_SUM_VALUE` -> `PRESERVE_ACCUMULATED_VALUE`

cgra/test/CgraWithContextSwitchRTL_test.py
```diff
@@ -874,9 +874,21 @@ def sim_fir_return(cmdline_opts, mem_access_is_combinational):
...
+      
+      # Terminate double time to make sure all registers are cleared
+      [
+        IntraCgraPktType(0, 0, payload = CgraPayloadType(CMD_TERMINATE)),
+        IntraCgraPktType(0, 1, payload = CgraPayloadType(CMD_TERMINATE)),
+        IntraCgraPktType(0, 4, payload = CgraPayloadType(CMD_TERMINATE)),
+        IntraCgraPktType(0, 5, payload = CgraPayloadType(CMD_TERMINATE)),
+        IntraCgraPktType(0, 8, payload = CgraPayloadType(CMD_TERMINATE)),
+        IntraCgraPktType(0, 9, payload = CgraPayloadType(CMD_TERMINATE))
+      ],
```
tancheng: You just want to make the clock tick a few cycles, right? Why immediate resume doesn't work?

mem/const/ConstQueueDynamicRTL.py
```diff
@@ -72,7 +72,7 @@ def load_const():
     @update_ff
     def update_wr_cur():
       not_full = (s.wr_cur < const_mem_size)
-      if s.reset | s.clear:
+      if s.reset:
```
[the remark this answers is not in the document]
tancheng: Ah, sorry, you are right. I was talking about `rd_cur`. (maybe when we switch kernel, we have to make this `| s.clear`, as two kernels would have different consts).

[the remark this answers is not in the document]
yyan7223: `if s.reset | s.clear:` is now applied to both `rd_cur` and `wr_cur`, and I reload the const in `CgraWithContextSwitchRTL_test.py`

yyan7223:
> We can merge this PR now, right? @yyan7223

Yes we can

PR #218, opened 2025-11-03
[CleanUp] Clean up all redundant arguments in constructors.

BenkangPeng:
This PR removes the redundant arguments in constructor of modules under folders `controller/` `fu/` `mem/` `noc/` `rf/` `tile/`.

### DIFF
--- a/cgra/CgraRTL.py
+++ b/cgra/CgraRTL.py
@@ -94,7 +94,7 @@ class CgraRTL(Component):
     s.send_data_on_boundary_west  = [SendIfcRTL(DataType) for _ in range(height)]
 
     # Components
-    s.tile = [TileRTL(CtrlPktType, CgraPayloadType,
+    s.tile = [TileRTL(CtrlPktType,
                       ctrl_mem_size,
                       data_mem_size_global, num_ctrl,
                       total_steps, 4, 2, s.num_mesh_ports,
--- a/cgra/CgraTemplateRTL.py
+++ b/cgra/CgraTemplateRTL.py
@@ -81,7 +81,7 @@ class CgraTemplateRTL(Component):
     # s.send_data_on_boundary_west  = [SendIfcRTL(DataType) for _ in range(height)]
 
     # Components
-    s.tile = [TileRTL(CtrlPktType, CgraPayloadType,
+    s.tile = [TileRTL(CtrlPktType,
                       ctrl_mem_size,
                       data_mem_size_global, num_ctrl,
                       total_steps, 4, 2, s.num_mesh_ports,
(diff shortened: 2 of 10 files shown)


cgra/CgraRTL.py
```diff
@@ -94,9 +94,7 @@ def construct(s, CgraPayloadType,
     s.send_data_on_boundary_west  = [SendIfcRTL(DataType) for _ in range(height)]
 
     # Components
-    s.tile = [TileRTL(DataType, PredicateType, CtrlPktType,
-                      CgraPayloadType, CtrlSignalType,
-                      data_bitwidth,
+    s.tile = [TileRTL(CtrlPktType, CgraPayloadType,
```
tancheng: Can we pass in `IntraCgraPktType` here w/o `CgraPayloadType`? as `CgraPayloadType` is part of the `IntraCgraPktType`.

tancheng (commented):
Most failures are just the number of arguments issue, should be easy for you to fix, just need a bit more time.

You can run `pytest .` in the VectorCGRA directory to see whether all can pass or not. So no need to always rely on Github Workflow.

BenkangPeng:
> Excellent work @BenkangPeng~! Any other modules left for the arg cleanup in your opinion?

I think that's all, because I cleaned up the parameters level by level, and then file by file.

tancheng:
Hi @BenkangPeng, if you search for `"IntraCgraPktType, CgraPayloadType"`, https://github.com/search?q=repo%3Atancheng%2FVectorCGRA+%22IntraCgraPktType%2C+CgraPayloadType%22&type=code
many places still pass in all types, for the unnecessary argument in test harness, we can avoid passing them in: https://github.com/tancheng/VectorCGRA/blob/15d120cca92d9f1598cf79898e8af6af3393b90c/multi_cgra/test/MeshMultiCgraRTL_test.py#L48-L68

Moreover, again, the Payload type can be derived from the intra-cgra-pkt: https://github.com/tancheng/VectorCGRA/blob/15d120cca92d9f1598cf79898e8af6af3393b90c/lib/messages.py#L314

Can you help to cleanup them ❤️  Thanks a lot~!

tancheng:
> > Hi @BenkangPeng, if you search for `"IntraCgraPktType, CgraPayloadType"`, https://github.com/search?q=repo%3Atancheng%2FVectorCGRA+%22IntraCgraPktType%2C+CgraPayloadType%22&type=code many places still pass in all types, for the unnecessary argument in test harness, we can avoid passing them in:
> > https://github.com/tancheng/VectorCGRA/blob/15d120cca92d9f1598cf79898e8af6af3393b90c/multi_cgra/test/MeshMultiCgraRTL_test.py#L48-L68
> > 
> > Moreover, again, the Payload type can be derived from the intra-cgra-pkt:
> > https://github.com/tancheng/VectorCGRA/blob/15d120cca92d9f1598cf79898e8af6af3393b90c/lib/messages.py#L314
> > 
> > Can you help to cleanup them ❤️ Thanks a lot~!
> 
> For all the test files, I don't remove any redundant args in constructors since the `TestHarness` class is redefined in different test cases without any reuse. Can simplify it if necessary.

Yeah, I understand that. Let's try to simplify them. As long as an arg is not used, let's remove it and its corresponding declarations in callers. Thanks~!

PR #222, opened 2025-11-08
Give a buffer before controller xbar to alleviate critical path

tancheng:
Leverage the xbar with buffer on input_unit.


zsjiang99 (approved):
Approve~

PR #230, opened 2025-11-23
[Refactor] Refactors reusable modules and functions while removing redundant code.

BenkangPeng:
This PR refactors reusable modules and functions while removing redundant code.

All the reusable modules and functions are placed into `lib/util/cgra`.

### DIFF
--- a/cgra/test/arch.yaml
+++ b/cgra/test/arch.yaml
@@ -3,8 +3,8 @@ multi_cgra_defaults:
   columns: 1
 
 cgra_defaults:
-  rows: 4
-  columns: 4
+  rows: 2
+  columns: 2
   configMemSize: 8
 
 tile_defaults:


cgra/test/CgraTemplateRTL_test.py
```diff
@@ -32,6 +32,10 @@
 from ...lib.messages import *
 from ...lib.opt_type import *
 from ...lib.util.common import *
+from ...lib.cgra.Tile import Tile
```
tancheng: move the `cgra` folder into `util`?

multi_cgra/test/MeshMultiCgraTemplateRTL_test.py
tancheng: `multiCgraParam` is `None` or not? I thought we would have an `arch.yaml` for this test, no?

cgra/test/CgraTemplateRTL_test.py
BenkangPeng: OK.

multi_cgra/test/MeshMultiCgraTemplateRTL_test.py
```diff
@@ -152,7 +152,10 @@ def run_sim(test_harness, max_cycles = 200):
   test_harness.sim_tick()
 
 
-def test_mesh_multi_cgra_universal(cmdline_opts, multiCgraParam = None):
+def test_mesh_multi_cgra_universal(cmdline_opts):
```
tancheng: I think your current param looks good: https://github.com/tancheng/VectorCGRA/blob/04b45159f4b69b840d302f31e453d33381c577e9/multi_cgra/test/MeshMultiCgraTemplateRTLScale_test.py#L10

i.e.,
```
def test_mesh_multi_cgra_universal(cmdline_opts, multiCgraParam = None):

if multiCgraParam == None:
  arch_file = os.path.join(os.path.dirname(__file__), "arch.yaml")
  parser = Parser(arch_file)
  multiCgraParam = parser.parse_multi_cgra_param()
```

How does this sound?

cgra/test/CgraTemplateRTL_test.py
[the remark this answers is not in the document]
BenkangPeng: Sorry, but what does `a default yaml` mean? A default `multiCgraParam` with size of `2x2x2x2` ? Or the file `arch.yaml` is optional and we should handle the case when `arch.yaml` doesn't exist?

multi_cgra/test/MeshMultiCgraTemplateRTL_test.py
```diff
@@ -152,16 +152,18 @@ def run_sim(test_harness, max_cycles = 200):
   test_harness.sim_tick()
 
 
-def test_mesh_multi_cgra_universal(cmdline_opts):
-  arch_file = os.path.join(os.path.dirname(__file__), "arch.yaml")
-  parser = Parser(arch_file)
-  multiCgraParam = parser.parse_multi_cgra_param()
+def test_mesh_multi_cgra_universal(cmdline_opts, multiCgraParam = None):
```
tancheng: Instead of `multiCgraParam`, can we have:
```
def test_mesh_multi_cgra_universal(cmdline_opts, arch_yaml_path = "arch.yaml"):
```

I think this would be easier to use, user no need to parse the yaml before calling this func then.

cgra/test/CgraTemplateRTL_test.py
```diff
@@ -103,14 +105,22 @@ def line_trace(s):
     return s.dut.line_trace()
 
 def test_cgra_universal(cmdline_opts, paramCGRA = None):
```
tancheng:
```
def test_cgra_universal(cmdline_opts, arch_yaml_path = "arch.yaml"):
```

cgra/test/arch.yaml
```diff
@@ -0,0 +1,12 @@
...
+  rows: 1
+  columns: 1
+
+cgra_defaults:
+  rows: 4
+  columns: 4
+  configMemSize: 8
+
+tile_defaults:
+  num_registers: 128
```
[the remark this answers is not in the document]
BenkangPeng: > Why is `num_registers: 128`? It is a bit too large though still works, but i am wondering why you provide such a number?

I just see it here.

https://github.com/coredac/dataflow/blob/4f815fc86020a60aad14089b1cf0d05288a15b97/test/arch_spec/arch_spec_example.yaml#L20-L21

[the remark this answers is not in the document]
tancheng: Let's use 16 in this PR.

PR #232, opened 2025-11-28
Update CrossbarRTL_test.py

EveandBob:
When testing with num_inports = 3 and num_outports = 4, I found that dut.send_data[4].rdy was always 0. After debugging, I discovered that CrossbarRTL_test.py needs to be modified.


tancheng:
Thanks for the fix~!

PR #247, opened 2026-01-06
Enable prototype RetVoid RTL

ShangkunLi:
In this pr, we enable a prototype RetVoid RTL:
- After receiving the trigger value, wait 100 cycles before sending the `CMD_COMPLETE` signal to the control memory
- The RetVoid Op is triggered by the predicate bit of the corresponding trigger value

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -111,8 +111,10 @@ jobs:
         pytest ../mem/data/test/DataMemControllerRTL_test.py -xvs --test-verilog --dump-vtb --dump-vcd
         # MeshMultiCgraTemplateRTL Test
         pytest ../multi_cgra/test/MeshMultiCgraTemplateRTL_test.py -xvs --test-verilog --dump-vtb --dump-vcd
-        # Generate .v file with explicit file & module names.
+        # Generate FIR scalar .v file with explicit file & module names.
         pytest ../multi_cgra/test/MeshMultiCgraRTL_test.py::test_multi_CGRA_fir_scalar_translation -xvs --test-verilog --dump-vtb --dump-vcd
+        # Generate FIR vector global reduce .v file with explicit file & module names.
+        pytest ../multi_cgra/test/MeshMultiCgraRTL_test.py::test_multi_CGRA_fir_vector_global_reduce_translation -xvs --test-verilog --dump-vtb --dump-vcd
 
     - name: Test synthesis for CGRA template
       run: |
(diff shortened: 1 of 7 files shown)


fu/single/RetRTL.py
```diff
@@ -100,9 +126,51 @@ def update_already_done():
...
             s.send_to_ctrl_mem.rdy:
           s.already_done <<= 1
+        elif s.recv_opt.val & \
+             (s.recv_opt.msg.operation == OPT_RET_VOID) & \
+              ~s.already_done & \
+              s.is_counting & \
+              (s.delay_counter == CounterType(kDelayCycles)) & \
+              s.send_to_ctrl_mem.val & \
+              s.send_to_ctrl_mem.rdy:
+          s.already_done <<= 1
```
tancheng: I think we have to make `s.already_done` be `s.already_done = [Wire(1) for _ in range(ctrl_mem_size)]` instead of `s.already_done = Wire(1)`, as there might be multiple `return` on same tile.

Or we can force the mapper avoid mapping multiple `return` on same tile, which one you prefer?

ShangkunLi: I think we can map multiple `return` on the same tile. It sounds more general.

tancheng: Makes sense. I think besides the [operation](https://github.com/tancheng/VectorCGRA/blob/517b6f9d835fab52a8b3608a069088dc2595ce77/lib/messages.py#L146), each config should also contain its `index` in the config memory.
 - It helps locate which `return` is done
 - When sending `COMPLETE` back to CPU, the `index` can be used to identify which task is returned. (CPU knows the mapping info).

tancheng: After adding new field, https://github.com/tancheng/VectorCGRA/blob/879941d5440127758c9fe2e88b15d0ded38f4e7a/fu/single/test/RetRTL_test.py#L94-L96
The test signal should also add the corresponding value:
```
 src_opt = [CtrlType(OPT_RET, 0, [FuInType(1), FuInType(0)]), 
            CtrlType(OPT_RET, 1, [FuInType(1), FuInType(0)]), 
            CtrlType(OPT_RET, 2, [FuInType(1), FuInType(0)])] 
```

Yes, all places need this change. However, a better way to enable this might be providing an outport in https://github.com/tancheng/VectorCGRA/blob/879941d5440127758c9fe2e88b15d0ded38f4e7a/tile/TileRTL.py#L108
(similar to [this](https://github.com/tancheng/VectorCGRA/blob/879941d5440127758c9fe2e88b15d0ded38f4e7a/mem/ctrl/CtrlMemDynamicRTL.py#L78)) and connect that to a new inport in the https://github.com/tancheng/VectorCGRA/blob/879941d5440127758c9fe2e88b15d0ded38f4e7a/tile/TileRTL.py#L84

ShangkunLi: Added `ctrl_addr_inport` in each FU.

PR #248, opened 2026-01-07
Add FIR vector global reduce SV test environment and fix Xprop in the CGRA RTL.

rp15:


multi_cgra/test/sv_test/header_fir_vector_global_reduce.sv
```diff
@@ -0,0 +1,358 @@
+function automatic IntraCgraPacket_4_2x2_16_8_2_CgraPayload__d294fd7ecd3c5b69 make_intra_cgra_pkt
```
tancheng: Can we avoid `d294fd7ecd3c5b69` by leveraging explicit name?

multi_cgra/test/sv_test/header_fir_vector_global_reduce.sv
```diff
@@ -0,0 +1,358 @@
...
+(
+  input logic [4:0] src,
+  input logic [4:0] dst,
+  input logic [4:0] cmd,
+  input logic [31:0] data_payload,
+  input logic       data_predicate,
+  input logic [6:0] data_addr,
+  input logic [6:0] ctrl_operation
+);
+  IntraCgraPacket_4_2x2_16_8_2_CgraPayload__d294fd7ecd3c5b69 pkt;
```
tancheng: ditto

multi_cgra/test/sv_test/header_fir_vector_global_reduce.sv
tancheng: Which part of this .sv can be used by riscv core? Or this file is just a verilog version of pymtl testbench?

tancheng (commented):
Also include the test inside https://github.com/tancheng/VectorCGRA/blob/master/.github/workflows/python-package.yml?

And update test/sv_test/README.md?

multi_cgra/test/sv_test/header_fir_vector_global_reduce.sv
rp15: I don't think we can control struct names in the same way as module/file names.

tancheng: So it will always be the same name?

multi_cgra/test/sv_test/header_fir_vector_global_reduce.sv
rp15: The integrated tb doesn't use this header, it uses only the .hex file generated by this header in a prior CGRA-only tb run.

multi_cgra/test/sv_test/header_fir_vector_global_reduce.sv
rp15: As long as the underlying RTL parameters are the same. That is why we have different versions for the scalar vs vector FIR tbs; the payload data is 32-bit wide in the former and 64 in the latter. I agree that it would be better to be more generic, but if the RTL bit widths change, we have more parameters to adjust by hand anyway, unfortunately.

multi_cgra/test/sv_test/README.md
```diff
@@ -5,7 +5,7 @@ They have two roles;
...
+1) is accomplished by header*.sv's functions (make_intra_cgra_pkt, make_intra_cgra_config_pkt, make_intra_cgra_config_pkt_w_data, unpack_pkt) that create the same packets as implemented by the by the IntraCgraPktType data types in preload_data and src_opt_pkt in MeshMultiCgraRTL_test.py. 2) is also accomplished by the same functions; in their second part, they print the necessary RISC-V Imem hex code into a file for the CPU to load the shared Dmem with the appropriate config packets (Imem init: https://github.com/coredac/accelerator_soc/blob/ed4e6f05d0ad88b62e347557a188ffca0dac3343/src/imem_rom.sv#L10, integration tb: https://github.com/coredac/accelerator_soc/blob/ed4e6f05d0ad88b62e347557a188ffca0dac3343/verif/acc_soc_tb.sv#L12).
```
tancheng: ([Imem init](https://github.com/coredac/accelerator_soc/blob/ed4e6f05d0ad88b62e347557a188ffca0dac3343/src/imem_rom.sv#L10), [integration tb](https://github.com/coredac/accelerator_soc/blob/ed4e6f05d0ad88b62e347557a188ffca0dac3343/verif/acc_soc_tb.sv#L12))

tancheng: []+()

PR #250, opened 2026-01-10
Hex code generation

rp15:
Enabled RISC-V hex code generation along with the already existing CGRA SV verification for the FIR vector global reduce VectorCGRA SV test.


rp15:
> I thought the test was already enabled in previous PR. So this PR does some cleanup or bug fix (then why it passed before)?

Good question; this PR is adding capability to generate RISC-V hex code to be used with the [coredac accelerator SOC](https://github.com/coredac/accelerator_soc) test bench. Not changing any of the functionality of this CGRA-only tb.

tancheng (approved):

PR #255, opened 2026-01-27
Enhance DCU with leaf counter & relay/root counter ability

ShangkunLi:
Screenshot 2026-01-27 at 15 52 39
 As shown in the figure, each DCU has two modes:
`data-drive` mode: this is used to execute the leaf counter
`data-provide` mode: this is used to provide outer loop values during execution

In this pr:
- we enhance the `LoopCounterRTL` with the ability to support both this two modes
- we enhance both modes with a per-ctrl-addr related registers, so that we can temporally execute multiple leaf counters on the same DCU
- we introduce the `CMD_RESET_LEAF_COUNTER` to reset the leaf counter and `CMD_UPDATE_COUNTER_SHADOW_VALUE` to update the shadow values in the DCU

### DIFF
--- a/fu/single/LoopCounterRTL.py
+++ b/fu/single/LoopCounterRTL.py
@@ -121,7 +121,7 @@ class LoopCounterRTL(Fu):
             if ~s.already_done[addr]:
               s.send_to_ctrl_mem.val @= b1(1)
               s.send_to_ctrl_mem.msg @= s.CgraPayloadType(
-                CMD_COMPLETE, DataType(0, 0), 0, s.recv_opt.msg, addr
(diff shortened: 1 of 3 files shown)


ShangkunLi:
> Then `data-provide` can be renamed as `loop-delivery`; `data-drive` can be renamed as `loop-driven`. WDYT?

Good idea! I have modified the name of the loop counter mode.

I have also created an integrated test that uses the loop counter to execute the `fir` test on a `2x2` CGRA, I can make a pr for that after merging this pr.

tancheng (approved):

PR #257, opened 2026-02-05
Update tile in type(port xx)

ycc-21:
Update all TileInType(PORT_XX) in test.


tancheng:
Thanks for the help @ycc-21~! @ShangkunLi Plz take a look.

PR #263, opened 2026-02-14
Refactor Fu interfaces.

yyan7223:


fu/flexible/FlexibleFuRTL.py
```diff
@@ -22,6 +22,7 @@
 
 class FlexibleFuRTL(Component):
   def construct(s,
+                CtrlPktType,
                 DataType,
                 CtrlType,
```
tancheng: `CtrlPktType` is already passed, then do we still need `DataType` and `CtrlType`?

yyan7223: Yes, let me do it step by step.

PR #270, opened 2026-03-04
Scratch/ronjok/dynamic handshake

rp15:
Dynamically handshake rdy & val using an unbounded queue.


multi_cgra/test/sv_test/MeshMultiCgraRTL_2x2_fir_scalar_tb.v
```diff
@@ -25,151 +24,78 @@ module cgra_test
   int  PASS         = 'd0;
   time pass_time_of = 'd0;
 
+  logic [185-1:0] pkt_queue [$];
```
tancheng: what is the queue size and the bit-width of each item inside the queue? Can you add comment for that?

tancheng (approved):

PR #271, opened 2026-03-05
Scratch/ronjok/fir scalar 2x2 2x2 pymtl

rp15:
I am opening this pull request to discuss the necessary modifications to execute scalar FIR PyMTL simulation on 2x2-CGRA grid with 2x2 tiles in every CGRA, moving from a 2x2 CGRA with 4x4 tiles. I have already changed the cfg packets to have

- the correct indexing in CGRA0 (bottom left) (tiles 4, 5 are now tiles 2, 3) and
- move the old CGRA0 tiles 8, 9 to CGRA2 (top left) (CGRA0 tiles 8, 9 are now CGRA 2 tiles 0, 1).
- Also changed expected_complete_sink_out_pkg not to violate the new bit width, but I am not sure what src and dst values should be in this scenario.

Currently getting an error from pytest that simulation exceeds the 200-cycle limit. What else would I need to modify moving forward? Thank you!

### DIFF
--- a/multi_cgra/test/MeshMultiCgraRTL_test.py
+++ b/multi_cgra/test/MeshMultiCgraRTL_test.py
@@ -799,6 +799,497 @@ def initialize_test_harness(cmdline_opts,
       ctrl_steps_per_iter = 3
       ctrl_steps_total = 3
 
+  elif test_name == 'test_systolic_4x4_2x2':
+      updated_ctrl_steps = 3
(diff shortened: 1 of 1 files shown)


tancheng:
right, that's the dump trace in every module. or you can use python `print()` in the FU you want to check, e.g., https://github.com/tancheng/VectorCGRA/blob/4d4937ad8562f0797ec78313418fb774b124bba2/fu/single/MulRTL.py#L65

to see what is the value during simulation.

PR #276, opened 2026-03-26
New PyMTL tests.

rp15:
Adding

- 2x2-CGRA 2x2-tile FIR scalar (prev. 2x2, 4x4 with no inter-CGRA communication) and
- 4x4-CGRA 2x2-tile Systolic array (prev. 2x2, 2x2)

tests to the Mesh multi-CGRA suite.


rp15:
> How is this different from #271? Plz point me to the specific line.

Copied from the email: Good question, they are identical materially, except for the parameter name change in the new test (commit 31d872). Due to the name change or lack thereof, 271 does not pass the tests since it is not merged with the latest master. I want to keep 271’s side branch without merging it with the latest CGRA RTL changes to make sure we have a working SV model. 271 is obsolete and can be closed without merging. (Sorry about the duplicate requests. I was not aware of the planned RTL change when I started 271; I would not have opened it otherwise.)

PR #278, opened 2026-03-28
PR for https://github.com/coredac/dataflow/issues/216

### LINKED ISSUE #216: [P3] Const related operation won't proceed const iter during prologue
Const related operations, e.g., phi_const
 - Requires prologue to skip the non-const operand for the first time
 - May or may not consume the const queue and the arrived data on the input port, which needs manually ordering the const data inside the const queue: https://github.com/tancheng/VectorCGRA/blob/9c1d72e5abc2aa92fe9c2f700f5b19c456a96d61/fu/single/PhiRTL.py#L96-L100
 - This becomes more complicated when prologue is more than 1 cycle and multiple const related operations mapped on that tile

yyan7223:
Now multiple PHI_CONST PHI_START should be mapped on the same tile. @tancheng @ShangkunLi @Jackcuii

### DIFF
--- a/fu/single/test/PhiRTL_test.py
+++ b/fu/single/test/PhiRTL_test.py
@@ -23,13 +23,16 @@ class TestHarness(Component):
 
   def construct(s, FunctionUnit, IntraCgraPktType, DataType, CtrlType,
                 num_inports, num_outports, data_mem_size, src0_msgs,
-                src1_msgs, src_const, src_opt, sink_msgs):
+                src1_msgs, src_const, src_opt, sink_msgs, src_ctrl_addr = []):
 
     s.src_in0 = TestSrcRTL(DataType, src0_msgs)
     s.src_in1 = TestSrcRTL(DataType, src1_msgs)
     s.src_const = TestSrcRTL(DataType, src_const)
     s.src_opt = TestSrcRTL(CtrlType, src_opt)
     s.sink_out = TestSinkRTL(DataType, sink_msgs)
+    s.CtrlAddrType = IntraCgraPktType.get_field_type(kAttrPayload).get_field_type(kAttrCtrlAddr)
+    if src_ctrl_addr:
+      s.src_ctrl_addr = TestSrcRTL(s.CtrlAddrType, src_ctrl_addr) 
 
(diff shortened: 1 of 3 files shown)


fu/single/test/PhiRTL_test.py
```diff
@@ -109,17 +115,22 @@ def test_Phi_start():
...
+  src_const = []
+               # Assumes that two PHI_START mapped on the same tile and iterates for twice.
   src_opt =   [CtrlType(OPT_PHI_START, pickRegister),
+               CtrlType(OPT_PHI_START, pickRegister),
                CtrlType(OPT_PHI_START, pickRegister),
                CtrlType(OPT_PHI_START, pickRegister)]
-
-  sink_out = [DataType(2, 1), DataType(5, 1), DataType(2, 1)]
+  src_ctrl_addr = [CtrlAddrType(0), CtrlAddrType(1), CtrlAddrType(0), CtrlAddrType(1)]
+  sink_out = [DataType(2, 1), DataType(3, 1), DataType(8, 1), DataType(3, 1)]
```
tancheng, line 130: The second output is `DataType(3, 1)`, however, the input is `DataType(3, 0)` on LHS.

I can't clearly remember my motivation to make the `phi_start`'s first time **NOT** rely on LHS' `predicate`: https://github.com/tancheng/VectorCGRA/blob/812915dd6f34e05bcc6f9bcbf00c24512cf1ff68/fu/single/PhiRTL.py#L88

@Jackcuii @ShangkunLi did you see any case `phi_start` first time need to output `predicate=0`? From kernel perspective, first time `phi_start` must output predicate of 1, is this reasonable?

tancheng (approved):
Thanks @yyan7223, looks perfect to me~!

fu/single/test/PhiRTL_test.py
tancheng: @ShangkunLi Can you please chime in?

ShangkunLi: We do have kernels that need the `phi_start` to output `predicate = 0` the first time; the `spmv` kernel is the case.

And I think it is not that necessary to introduce `phi_start` as discussed in https://github.com/coredac/dataflow/pull/295#issuecomment-4100518342

Issue #281, opened 2026-04-08
[P0] Simulator and rtl comparison (latency gap) debug


n0thingNoob:
In the histogram kernel, we found an issue, but we don't know why it happens.
So in the program we used(The red blocks are two key ops where the issue happened. ):
<img width="552" height="821" alt="Image" src="https://github.com/user-attachments/assets/6937237b-462e-4d1e-88da-c838b79b346f" />

<img width="876" height="209" alt="Image" src="https://github.com/user-attachments/assets/f57b1a96-5496-4267-8b4e-154c0df1ec30" />
The NAH operation in the red block is a routing operation, corresponding to the CTRL_MOV in the previous assembly code. It is expected to take only one cycle, but in the RTL trace, it appears to stall for an extra cycle and completes in the second cycle instead. Since I am still learning how to read the RTL logs, I have attached a CSV file with the key operations for further inspection.

[tile_1_0_chain_126_134_key_only.csv](https://github.com/user-attachments/files/26582446/tile_1_0_chain_126_134_key_only.csv)

Here is a visualized timeseries graph for the steady cycle in this screenshot. The red block means stalls. Blue and green cycle means it is completed. 
<img width="996" height="510" alt="Image" src="https://github.com/user-attachments/assets/1c7f3780-c756-4d09-b418-55b586e2d08e" />
Here is the visualized graph for simulator:

<img width="949" height="478" alt="Image" src="https://github.com/user-attachments/assets/776c9f89-ea25-435b-a61d-f6f65be000b8" />

Let me also attach the log for both:

[histogram.json.log](https://github.com/user-attachments/files/26582484/histogram.json.log) (Simulator log)

[trace_histogram_4x4_Mesh.txt](https://github.com/user-attachments/files/26582487/trace_histogram_4x4_Mesh.txt) (RTL log)

n0thingNoob:
@tancheng @yyan7223

tancheng:
Thanks for filing the issue @n0thingNoob, are you able to provide the reproduce commands? e.g., a pymtl test?

@yyan7223 can you help to investigate why there is an additional cycle (doing nothing) between `grant` (on `tile[1,1]`) and `phi` (on `tile [1,0]`)?

n0thingNoob:
The histogram pytest file is in (https://github.com/tancheng/VectorCGRA/tree/fulfill-kernel-legacy/cgra/test)

[...]

PR #282, opened 2026-04-12
Fixes two-cycles bug of NAH operation.

yyan7223:
Have not tested whether the bug can be fixed by this PR.

### DIFF
--- a/mem/register_cluster/RegisterClusterRTL.py
+++ b/mem/register_cluster/RegisterClusterRTL.py
@@ -90,7 +90,8 @@ class RegisterClusterRTL(Component):
             (s.reg_bank[i].send_data.val & reg_towards_fu)
         s.reg_bank[i].send_data.rdy @= s.send_data_to_fu[i].rdy
 
-        s.recv_data_from_routing_crossbar[i].rdy @= (s.inport_opt.operation == OPT_NAH) | (s.inport_opt.fu_in[i] == 0) | s.send_data_to_fu[i].rdy
+        s.recv_data_from_routing_crossbar[i].rdy @= ((s.inport_opt.write_reg_from[i] == PORT_ROUTING_CROSSBAR) \
+                & (s.inport_opt.operation == OPT_NAH)) | s.send_data_to_fu[i].rdy
         s.recv_data_from_fu_crossbar[i].rdy @= 1
         s.recv_data_from_const[i].rdy @= 1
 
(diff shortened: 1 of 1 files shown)


mem/register_cluster/RegisterClusterRTL.py
```diff
@@ -90,7 +90,7 @@ def update_msgs_signals():
             (s.reg_bank[i].send_data.val & reg_towards_fu)
         s.reg_bank[i].send_data.rdy @= s.send_data_to_fu[i].rdy
 
-        s.recv_data_from_routing_crossbar[i].rdy @= s.send_data_to_fu[i].rdy
+        s.recv_data_from_routing_crossbar[i].rdy @= (s.inport_opt.operation == OPT_NAH) | (s.inport_opt.fu_in[i] == 0) | s.send_data_to_fu[i].rdy
```
tancheng: This logic resolves the issue, specifically for the cases:
 - Only towards register cluster -> can always write
 - Towards both register and FU -> depend on `s.send_data_to_fu[i].rdy`
 - Only towards FU -> depend on `s.send_data_to_fu[i].rdy`
 - But what about towards nothing?
   - i.e., my concern is whether this would over-consume some pending data on the inport. The top of fifo has a data valid, but it would be used/consumed in the future, instead of current instruction. Current instruction is `OPT_NAH`, so the `s.recv_data_from_routing_crossbar[i].rdy` is `1`.

@yyan7223 WDYT?

tancheng: I feel we need to have a `routing_xbar_towards_reg[i] @= write_reg_from[i] == PORT_ROUTING_CROSSBAR`

```
s.recv_data_from_routing_crossbar[i].rdy @= (routing_xbar_towards_reg[i] & (s.inport_opt.operation == OPT_NAH)) | s.send_data_to_fu[i].rdy
```
in other words:
```
s.recv_data_from_routing_crossbar[i].rdy @= ((write_reg_from[i] == PORT_ROUTING_CROSSBAR) & (s.inport_opt.operation == OPT_NAH)) | s.send_data_to_fu[i].rdy
```

@Jackcuii can you also try this and see whether it helps https://github.com/tancheng/VectorCGRA/issues/281#issuecomment-4231999999? I don't quite understand the two problems you described though.

Jackcuii: Can we have a quick sync？ It is a bit hard to describe clearly with text.

yyan7223:
```
s.recv_data_from_routing_crossbar[i].rdy @= ((write_reg_from[i] == PORT_ROUTING_CROSSBAR) & (s.inport_opt.operation == OPT_NAH)) | s.send_data_to_fu[i].rdy
```
I test it on some imagined scenarios and it looks good to me.

tancheng: then update the code~?

tancheng (approved):

PR #285, opened 2026-04-15
Fixes the bug of unexpected stall cycles during prologue.

yyan7223:
@Jackcuii https://github.com/tancheng/VectorCGRA/issues/281#issuecomment-4231999999 should have been fixed in this PR, can you check whether it works on your side?

@tancheng `Replacing the operation with OPT_NAH during the prologue` is now implemented in the ctrl mem. I have also further restricted some control signals in the crossbar to toggle only after the prologue, and added a test to confirm that the unexpected stall cycles during the prologue are fixed. Please review.

### DIFF
--- a/mem/ctrl/CtrlMemDynamicRTL.py
+++ b/mem/ctrl/CtrlMemDynamicRTL.py
@@ -88,7 +88,6 @@ class CtrlMemDynamicRTL(Component):
         [[Wire(PrologueCountType) for _ in range(num_routing_xbar_inports)] for _ in range(ctrl_mem_size)]
 
     # Connections.
-    s.send_ctrl.msg //= s.reg_file.rdata[0]
     s.recv_pkt_from_controller //= s.recv_pkt_from_controller_queue.recv
     s.recv_from_element //= s.recv_from_element_queue.recv
 
@@ -206,6 +205,24 @@ class CtrlMemDynamicRTL(Component):
       if s.recv_pkt_from_controller_queue.send.val & \
           (s.recv_pkt_from_controller_queue.send.msg.payload.cmd == CMD_TERMINATE):
         s.send_ctrl.val @= b1(0)
+      
+      for i in range(num_fu_inports):
+        s.send_ctrl.msg.fu_in[i]            @= s.reg_file.rdata[0].fu_in[i]
+        s.send_ctrl.msg.write_reg_from[i]   @= s.reg_file.rdata[0].write_reg_from[i]
+        s.send_ctrl.msg.write_reg_idx[i]    @= s.reg_file.rdata[0].write_reg_idx[i]
+        s.send_ctrl.msg.read_reg_towards[i] @= s.reg_file.rdata[0].read_reg_towards[i]
+        s.send_ctrl.msg.read_reg_idx[i]     @= s.reg_file.rdata[0].read_reg_idx[i]
+      for i in range(num_routing_outports):
+        s.send_ctrl.msg.routing_xbar_outport[i] @= s.reg_file.rdata[0].routing_xbar_outport[i]
+        s.send_ctrl.msg.fu_xbar_outport[i]      @= s.reg_file.rdata[0].fu_xbar_outport[i]
+      s.send_ctrl.msg.vector_factor_power @= s.reg_file.rdata[0].vector_factor_power
+      s.send_ctrl.msg.is_last_ctrl        @= s.reg_file.rdata[0].is_last_ctrl
+      # Sets each FU's op code as NAH when prologue execution is not completed.
+      # As they are supposed to do nothing during that prologue cycles.
+      if s.prologue_count_outport_fu != 0:
+        s.send_ctrl.msg.operation @= OPT_NAH
+      else:
+        s.send_ctrl.msg.operation @= s.reg_file.rdata[0].operation
 
     @update_ff
     def update_whether_we_can_iterate_ctrl():
--- a/mem/ctrl/CtrlMemDynamicRTL.py
+++ b/mem/ctrl/CtrlMemDynamicRTL.py
@@ -192,7 +192,7 @@ class CtrlMemDynamicRTL(Component):
             s.send_pkt_to_controller.val @= 1
 
     @update
-    def update_send_ctrl():
+    def update_send_ctrl_val():
       s.send_ctrl.val @= 0
       if s.start_iterate_ctrl == b1(1):
         if s.sent_complete:
@@ -205,7 +205,9 @@ class CtrlMemDynamicRTL(Component):
       if s.recv_pkt_from_controller_queue.send.val & \
           (s.recv_pkt_from_controller_queue.send.msg.payload.cmd == CMD_TERMINATE):
         s.send_ctrl.val @= b1(0)
-      
+
+    @update
+    def update_send_ctrl_msg():
       for i in range(num_fu_inports):
         s.send_ctrl.msg.fu_in[i]            @= s.reg_file.rdata[0].fu_in[i]
         s.send_ctrl.msg.write_reg_from[i]   @= s.reg_file.rdata[0].write_reg_from[i]
@@ -217,8 +219,8 @@ class CtrlMemDynamicRTL(Component):
         s.send_ctrl.msg.fu_xbar_outport[i]      @= s.reg_file.rdata[0].fu_xbar_outport[i]
       s.send_ctrl.msg.vector_factor_power @= s.reg_file.rdata[0].vector_factor_power
       s.send_ctrl.msg.is_last_ctrl        @= s.reg_file.rdata[0].is_last_ctrl
-      # Sets each FU's op code as NAH when prologue execution is not completed.
-      # As they are supposed to do nothing during that prologue cycles.
+      # Sets each FU's op code as NAH when prologue execution has not completed.
+      # As FU is supposed to do nothing during prologue.
       if s.prologue_count_outport_fu != 0:
         s.send_ctrl.msg.operation @= OPT_NAH
       else:
--- a/noc/CrossbarRTL.py
+++ b/noc/CrossbarRTL.py
@@ -237,7 +237,8 @@ class CrossbarRTL(Component):
         # (i.e., i >= num_inports) go to the FU's inports. In other words, we skip
         # the rdy checking on the FU's inports (connecting from crossbar_outport) if
         # the compute is already completed.
-        if (s.in_dir[i] > 0) & \
+        # Those conditions are effective only after prologue.
+        if (s.in_dir[i] > 0) & (~s.prologue_allowing_vector[i]) & \
            (~s.compute_done | (i < outport_towards_local_base_id)):
           s.send_rdy_vector[i] @= s.send_data[i].rdy
         else:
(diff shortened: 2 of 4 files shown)


mem/ctrl/CtrlMemDynamicRTL.py
```diff
@@ -206,6 +205,24 @@ def update_send_ctrl():
...
+        s.send_ctrl.msg.routing_xbar_outport[i] @= s.reg_file.rdata[0].routing_xbar_outport[i]
+        s.send_ctrl.msg.fu_xbar_outport[i]      @= s.reg_file.rdata[0].fu_xbar_outport[i]
+      s.send_ctrl.msg.vector_factor_power @= s.reg_file.rdata[0].vector_factor_power
+      s.send_ctrl.msg.is_last_ctrl        @= s.reg_file.rdata[0].is_last_ctrl
+      # Sets each FU's op code as NAH when prologue execution is not completed.
+      # As they are supposed to do nothing during that prologue cycles.
+      if s.prologue_count_outport_fu != 0:
+        s.send_ctrl.msg.operation @= OPT_NAH
+      else:
+        s.send_ctrl.msg.operation @= s.reg_file.rdata[0].operation
```
tancheng, line 227: Can we just do?
```
s.send_ctrl.msg @= s.reg_file.rdata[0]
.
.
.
if s.prologue_count_outport_fu != 0:
        s.send_ctrl.msg.operation @= OPT_NAH
```
i.e., keep original line 91, overwrite it when `prologue_count_outport_fu != 0`?

noc/CrossbarRTL.py
tancheng: Change "The `num_inports` indicates" to "The `outport_towards_local_base_id` indicates".

noc/CrossbarRTL.py
```diff
@@ -237,7 +237,8 @@ def update_rdy_vector():
         # (i.e., i >= num_inports) go to the FU's inports. In other words, we skip
         # the rdy checking on the FU's inports (connecting from crossbar_outport) if
         # the compute is already completed.
-        if (s.in_dir[i] > 0) & \
+        # Those conditions are effective only after prologue.
```
tancheng: Elaborate a bit more?
"Those conditions are effective only after prologue. If it is still in prologue, .... (everything should be ready?)"

noc/CrossbarRTL.py
```diff
@@ -247,7 +248,7 @@ def update_rdy_vector():
     def update_valid_vector():
       s.recv_valid_vector @= 0
       for i in range(num_outports):
-        if s.in_dir[i] > 0:
+        if (s.in_dir[i] > 0) & (~s.prologue_allowing_vector[i]):
```
tancheng: We already have `s.recv_valid_or_prologue_allowing_vector[i] @= s.recv_valid_vector[i] | s.prologue_allowing_vector[i]`, do we really need this?

In other words, above `send_rdy_vector` is not enough? The main purpose of this PR is to allow xbar's data to be stored in register and proceed to the next opt when the current opt is `NAH`, right?

noc/CrossbarRTL.py
```diff
@@ -258,7 +259,7 @@ def update_recv_required_vector():
         s.recv_required_vector[i] @= 0
 
       for i in range(num_outports):
-        if s.in_dir[i] > 0:
+        if (s.in_dir[i] > 0) & (~s.prologue_allowing_vector[i]):
```
tancheng: Why prologue has impact on `recv_required_vector`?

noc/CrossbarRTL.py
```diff
@@ -268,7 +269,7 @@ def update_send_required_vector():
         s.send_required_vector[i] @= 0
 
       for i in range(num_outports):
-        if s.in_dir[i] > 0:
+        if (s.in_dir[i] > 0) & (~s.prologue_allowing_vector[i]):
```
tancheng: ditto

mem/ctrl/CtrlMemDynamicRTL.py
yyan7223: Probably we can't, please remind me if I made any mistakes.

Originally I tried to keep `s.send_ctrl.msg //= s.reg_file.rdata[0]` at line 91,
but modifing the `s.send_ctrl.msg.operation` in an update block incurs a `MultipleWriterError`. 
655a66dd-1bfe-4709-9c25-2b537ba228bf

I also tried to put `s.send_ctrl.msg @= s.reg_file.rdata[0]` in the update block, but generates `NoWriterError` below.
17dfcb80-0876-4a49-9f5c-d15e8bdfeda2

So that's why I do it like now, do you see any issues?

tancheng: you meant `s.send_ctrl.msg @= s.reg_file.rdata[0]` instead of `s.send_ctrl.msg //= s.reg_file.rdata[0]`?

I thought it should be fine as we do have overwrite cases: https://github.com/tancheng/VectorCGRA/blob/df450725c6c0ade6fd29325c6898d763e16d0e48/fu/single/MemUnitRTL.py#L138-L142

tancheng: oh, I meant to put `s.send_ctrl.msg @= s.reg_file.rdata[0]` around line 195 within this block.

noc/CrossbarRTL.py
yyan7223: This seems necessary, I think crossbar should not consume any values from `tile_in_channel` or somewhere else during prologue?
At least it is the truth in my testbench. OPT_GRT_PRED takes inputs from `tile_in_channel` after prologue, so we shouldn't allow any inputs mistakenly consumed by crossbar during crossbar's prologue.

noc/CrossbarRTL.py
yyan7223: The crossbar in prologue should do nothing, right? So why we still have `s.send_rdy_vector[i] @= s.send_data[i].rdy`?

So I simply just set `send_rdy_vector`, `s.recv_valid_vector`, `s.recv_required_vector`, and `s.send_required_vector` all to 1 during crossbar's prologue, so as to make sure crossbar is not blocked in any condition during prologue.

And yes, this PR is for this. But I also consider other scenario like both xbar and fu are prologued in the testcase:
https://github.com/yyan7223/VectorCGRA/blob/4cadfad60a83c72fb23582bef5aa3fdff59ec41e/tile/test/TileRTL_test.py#L640-L649

noc/CrossbarRTL.py
yyan7223: Same reason as [above](https://github.com/tancheng/VectorCGRA/pull/285#discussion_r3115007051)

mem/ctrl/CtrlMemDynamicRTL.py
yyan7223: 2c9e6e21-6c96-4811-86b9-27c5dc5f2582
This works for `CtrlMemDynamicRTL_test.py`, but fails for `TileRTL_test.py`.

It is fine in MemUnit possibly because `predicate` is a single signal while `oepration` is a bunch of signals.

tancheng: I don't think so..... What is the error msg for TileRTL_test.py?

noc/CrossbarRTL.py
tancheng: no? `OPT_GRT_PRED` consumes input from the register instead of `tile_in_channel`: https://github.com/tancheng/VectorCGRA/issues/281#issuecomment-4231999999

noc/CrossbarRTL.py
```diff
@@ -258,7 +259,7 @@ def update_recv_required_vector():
         s.recv_required_vector[i] @= 0
 
       for i in range(num_outports):
-        if s.in_dir[i] > 0:
+        if (s.in_dir[i] > 0) & (~s.prologue_allowing_vector[i]):
           s.recv_required_vector[s.in_dir_local[i]] @= 1
```
tancheng: I am also wondering if you really want to check, should you then check `s.prologue_allowing_vector[i]` or `s.prologue_allowing_vector[s.in_dir_local[i]]`...

noc/CrossbarRTL.py
tancheng: My original intention is to keep semantic correctness as much as possible. valid is valid, prologue is prologue, but you are trying to implicitly embed prologue into other signals. It is okay, then `recv_valid_or_prologue_allowing_vector` seems not necessary any more? And should we rename all signals to `xxx_or_in_prologue`?

noc/CrossbarRTL.py
yyan7223: I consider two scenarios in the test:

Scenario1: Takes inputs from tile_in_channel
https://github.com/yyan7223/VectorCGRA/blob/4cadfad60a83c72fb23582bef5aa3fdff59ec41e/tile/test/TileRTL_test.py#L640-L649

Scenario2: Takes inputs from register_cluster
https://github.com/yyan7223/VectorCGRA/blob/4cadfad60a83c72fb23582bef5aa3fdff59ec41e/tile/test/TileRTL_test.py#L676-L685

noc/CrossbarRTL.py
yyan7223: Current solution may not be the best. I'll take the time to thoroughly understand the CrossbarRTL, and then rework this part. 

But CrossbarRTL now has bugs as it cannot pass the `test_prologue_nah` in `TileRTL_test.py`.

noc/CrossbarRTL.py
yyan7223: Honestly I don't understand what does `allow` mean, so I simply just follow your code
https://github.com/yyan7223/VectorCGRA/blob/4cadfad60a83c72fb23582bef5aa3fdff59ec41e/noc/CrossbarRTL.py#L203-L208
It seems that `prologue_counter < prologue_count_wire` is still in the prologue stage? So I use `~s.prologue_allowing_vector[i]` to indicate the execution stage.

mem/ctrl/CtrlMemDynamicRTL.py
yyan7223: [error.txt](https://github.com/user-attachments/files/26926304/error.txt)

tancheng: `s.send_ctrl.msg @= s.reg_file.rdata[0]` is inside the update_block outside any `if`, right?

noc/CrossbarRTL.py
tancheng: I am a bit lost for the case this PR is fixing. The original issue is xbar prologue once, while FU prologue twice. Maybe we can discuss this tmr.

noc/CrossbarRTL.py
tancheng: `s.prologue_allowing_vector[i]` means the port is in prologue stage, so we allow it to be skipped. Maybe we need to rename them all. Maybe `in_prologue_vector[i]`?

noc/CrossbarRTL.py
```diff
@@ -199,22 +199,22 @@ def update_send_accepted_next():
...
       for i in range(num_outports):
         if s.in_dir[i] > 0:
           # Records whether the prologue steps have already been satisfied.
-          s.prologue_allowing_vector[i] @= \
+          s.during_prologue_allowing_vector[i] @= \
             (s.prologue_counter[s.ctrl_addr_inport][s.in_dir_local[i]] < \
              s.prologue_count_wire[s.ctrl_addr_inport][s.in_dir_local[i]])
         else:
-          s.prologue_allowing_vector[i] @= 1
+          s.during_prologue_allowing_vector[i] @= 0
```
tancheng, line 210: I think your changes make super sense. If not in prologue, whether to proceed should depend on recv_valid.

But would this fix can pass all tests?

tancheng (approved):

PR #287, opened 2026-05-03
Feature/4 port lsu upgrade

wendywanghx27:
Implement 4-port Load/Store architecture and diagonal fabric routing. Update default DFG JSON to 16-bit token routing to resolve deadlock.


tancheng:
Hi @wendywanghx27, can you please elaborate on the problem we are trying to fix here? And summarize the proposed fix?

wendywanghx27:
opened against upstream by mistake sorry!

PR #290, opened 2026-05-08
[Doc] Unify coordinate system of multi-cgra

BenkangPeng:
This PR adds documentation for the x–y coordinate system of Multi-CGRA, as the current codebase mixes multiple coordinate systems (e.g., row–column and x–y), making conversions inconvenient. The x–y coordinate system is considered more intuitive and easier to use, I think :-)


tancheng (approved):
Thanks for the cleanup~!

PR #293, opened 2026-06-02
[Feature] Implement DMA support

BenkangPeng:
Related issue: https://github.com/coredac/CGRA-SoC/issues/2

This PR introduces `CgraDmaRTL` which integrates the CGRA with a DMA engine, enabling direct memory transfers between external DRAM(don't implement now) and the CGRA's dataSPM.


mem/dma/DmaEngineRTL.py
```diff
@@ -0,0 +1,317 @@
...
+memory interface and the CGRA dataSPM.
+"""
+
+from pymtl3 import *
+
+# DMA Move In and Out
+# DMA_MVIN  : DRAM -> DMA Engine -> SPM
+# DMA_MVOUT : SPM -> DMA Engine -> DRAM
+DMA_MVIN  = 0
+DMA_MVOUT = 1
```
tancheng: Put these into https://github.com/tancheng/VectorCGRA/blob/master/lib/util/common.py?

mem/dma/DmaEngineRTL.py
```diff
@@ -0,0 +1,317 @@
...
+
+
+class DmaEngineRTL( Component ):
+  """
+  The DmaEngineRTL module is responsible for bulk data movement between an
+  external DRAM-like memory and the on-chip Scratchpad Memory (dataSPM).
+
+  It supports two main operations:
+  - DMA_MVIN:  DRAM -> DMA Engine -> SPM
```
tancheng, line 23: Rename to `DMA_INFEED` and `DMA_OUTFEED`?

mem/dma/DmaEngineRTL.py
```diff
@@ -0,0 +1,317 @@
...
+from pymtl3 import *
+
+# DMA Move In and Out
+# DMA_MVIN  : DRAM -> DMA Engine -> SPM
+# DMA_MVOUT : SPM -> DMA Engine -> DRAM
+DMA_MVIN  = 0
+DMA_MVOUT = 1
+
+
+class DmaEngineRTL( Component ):
```
tancheng, line 17: Would this also handle instructions/control signals delivery besides data?

mem/dma/DmaEngineRTL.py
```diff
@@ -0,0 +1,317 @@
...
+
+    OpcodeType   = mk_bits( 3 )
+    DramAddrType = mk_bits( dram_addr_nbits )
+    SpmAddrType  = mk_bits( spm_addr_nbits )
+    BytesType    = mk_bits( bytes_nbits )
+    TagType      = mk_bits( tag_nbits )
+    SpmDataType  = mk_bits( spm_data_nbits )
+    MemDataType  = mk_bits( mem_data_nbits )
+    # Byte mask for SPM write; 1 byte = 8 bits
+    SpmMaskType  = mk_bits( spm_data_nbits // 8 )
```
tancheng: Define CHAR_BIT = 8 inside https://github.com/tancheng/VectorCGRA/blob/master/lib/util/common.py and use it here: ` // CHAR_BIT`

mem/dma/DmaEngineRTL.py
```diff
@@ -0,0 +1,317 @@
...
+    s.dma_cmd_bytes     = InPort( BytesType )
+    s.dma_cmd_tag       = InPort( TagType )
+
+    s.dma_done_val      = OutPort()
+    s.dma_done_rdy      = InPort()
+    s.dma_done_tag      = OutPort( TagType )
+
+    # Abstract external memory interface
+    # Request to read from DRAM
+    s.mem_rd_req_val    = OutPort()
```
tancheng: change all `mem` to `dram`, to distinguish it from `sram`?

mem/dma/DmaEngineRTL.py
```diff
@@ -0,0 +1,317 @@
...
+    s.spm_dma_rrdy      = InPort()
+    s.spm_dma_raddr     = OutPort( SpmAddrType )
+
+    # Response from SPM
+    s.spm_dma_rresp_val  = InPort()
+    s.spm_dma_rresp_rdy  = OutPort()
+    s.spm_dma_rresp_data = InPort( SpmDataType )
+
+    # State machine definitions
+    StateType = mk_bits( 4 )
```
tancheng: 4 status needs to be defined inside https://github.com/tancheng/VectorCGRA/blob/master/lib/util/common.py

cgra/CgraDmaRTL.py
```diff
@@ -0,0 +1,227 @@
...
+    s.dma_cmd_bytes     = InPort(DmaBytesType) # The number of bytes to transfer.
+    s.dma_cmd_tag       = InPort(DmaTagType) # Doesn't use it now, but keep it for future use(e.g., distinguish different DMA commands).
+
+    s.dma_done_val      = OutPort()
+    s.dma_done_rdy      = InPort()
+    s.dma_done_tag      = OutPort(DmaTagType) # Must be same as the input `dma_cmd_tag`
+
+    s.mem_rd_req_val    = OutPort() # dma_read_request_valid
+    s.mem_rd_req_rdy    = InPort() # dma_read_request_ready
+    s.mem_rd_req_addr   = OutPort(DmaDramAddrType)
```
tancheng: Why can't we use the `RecvIfcRTL` and `SendIfcRTL` interfaces to connect the `DmaRTL`?

mem/dma/DmaEngineRTL.py
BenkangPeng: Fixed.

mem/dma/DmaEngineRTL.py
BenkangPeng: Hi @HobbitQia , WDYT? Keep same names with gemmini or change them?

mem/dma/DmaEngineRTL.py
BenkangPeng: Current implementation doesn't specify data or instructions, maybe it depends how the CGRA consumes the data transferred from dram?

mem/dma/DmaEngineRTL.py
BenkangPeng: Fixed.

mem/dma/DmaEngineRTL.py
BenkangPeng: Fixed.

mem/dma/DmaEngineRTL.py
BenkangPeng: Fixed.

cgra/CgraDmaRTL.py
BenkangPeng: Fixed.

mem/data/DataMemControllerRTL.py
```diff
@@ -1,451 +1,581 @@
...
-     Blocking vs. non-blocking should be configured/propagated here.
-   - Non-blocking:
-     - Immediate return data though it is not ready:
-       - Bank conflicted lower priority access.
-       - Remote accessed data.
-   - Blocking and non-blocking might be configurabled in a dynamic way.
-
-Author : Cheng Tan
-  Date : Aug 28, 2025
-"""
```
BenkangPeng: It's weird that GitHub shows the diff this way. I didn't delete the whole file and re-add it — I only made some partial changes. Seeing "+581 -451" makes it really hard to review...

mem/dma/DmaEngineRTL.py
HobbitQia: we can use our own names. Gemmini's name has nothing to do with this.

cgra/CgraTemplateRTL.py
```diff
@@ -190,6 +213,22 @@ def construct(s, CgraPayloadType,
...
+      s.data_mem.spm_dma_wdata //= s.spm_dma_wdata
+      s.data_mem.spm_dma_wmask //= s.spm_dma_wmask
+
+      # DMA_MVOUT: spm -> dma -> dram
+      s.data_mem.spm_dma_rval       //= s.spm_dma_rval
+      s.data_mem.spm_dma_rrdy       //= s.spm_dma_rrdy
+      s.data_mem.spm_dma_raddr      //= s.spm_dma_raddr
+      s.data_mem.spm_dma_rresp_val  //= s.spm_dma_rresp_val
+      s.data_mem.spm_dma_rresp_rdy  //= s.spm_dma_rresp_rdy
+      s.data_mem.spm_dma_rresp_data //= s.spm_dma_rresp_data
```
tancheng: As we discussed before, should we connect the dma to the controller (as intermediate interface/transition)? instead of directly connecting to data spm?

So then we can leverage https://github.com/tancheng/VectorCGRA/blob/44618d5cc28bddf4130170b9da237c6229d91431/controller/ControllerRTL.py#L138-L139 to decide the next location (e.g., local spam bank, remote spm)?

mem/data/DataMemControllerRTL.py
BenkangPeng: > It's weird that GitHub shows the diff this way. I didn't delete the whole file and re-add it — I only made some partial changes. Seeing "+581 -451" makes it really hard to review...

Solved in https://github.com/tancheng/VectorCGRA/pull/294.

HobbitQia:
Hi @tancheng @BenkangPeng , I summarized two direction of DMA design as below:

- Rely on data controller
    - DMA is added as a new client of the DataMemControllerRTL, where data in the DMA engine communicates directly with the DataMemControllerRTL, and the logic for multiplexing SPM ports is also implemented in that module.
        
        To initiate DMA, the CPU can send `dma_mvin` or `dma_mvout` to the CGRA, after which the controller activates the DMA  engine by sending start signals.
        
    - **Pros**: Keeps the controller clean; provides a faster path because data does not go through the controller.
    - **Cons**: Additional logic is required to feed DMA results into the control memory.
   
     image

- All in controller
    - All decoding logic is handled in the controller. The logic for handling the logic port should still reside in the DataMemControllerRTL, since the data memory should have its own port multiplexing logic.
        
        The logic of packeting should also be implemented in the controller module.
        
    - **Pros**: Unifies control and data memory within the controller (the controller is already connected to both control and data memory).
    - **Cons**: Introduces complex control logic in the controller; results in a slower path.
     image

I prefer the second method but I think there are still some logic should be written in DataMemControllerRTL. WDTY?

tancheng:
Hi @HobbitQia, option 2 looks good to me. Though I am not sure what logic should be additionally in `DataMemController`?

cgra/CgraDmaRTL.py
```diff
@@ -0,0 +1,217 @@
...
+    s.dma.spm_dma_waddr      //= s.cgra.spm_dma_waddr
+    s.dma.spm_dma_wdata      //= s.cgra.spm_dma_wdata
+    s.dma.spm_dma_wmask      //= s.cgra.spm_dma_wmask
+
+    s.dma.spm_dma_rval       //= s.cgra.spm_dma_rval
+    s.dma.spm_dma_rrdy       //= s.cgra.spm_dma_rrdy
+    s.dma.spm_dma_raddr      //= s.cgra.spm_dma_raddr
+    s.dma.spm_dma_rresp_val  //= s.cgra.spm_dma_rresp_val
+    s.dma.spm_dma_rresp_rdy  //= s.cgra.spm_dma_rresp_rdy
+    s.dma.spm_dma_rresp_data //= s.cgra.spm_dma_rresp_data
```
tancheng: All these will change if we go with @HobbitQia's option 2, right?

Moreover, can we use send/recv interfaces, and define msg (https://github.com/tancheng/VectorCGRA/blob/master/lib/messages.py) to encapsulate data, addr, or whatever needed as struct? So we don't need to declare so many ports, and explicitly connect each of them. This CGRA RTL shouldn't see these details, the data struct can be decomposed inside the submodule.

HobbitQia:
> Hi @HobbitQia, option 2 looks good to me. Though I am not sure what logic should be additionally in `DataMemController`?

I am thinking if we want enable the concurrent running of DMA and traditional load/store, then we need to multiplex the port of Data SPM and I think this logic can be implemented in `DataMemController`. However, we can also transform the data from DRAM into packets and use the command like `CMD_STORE_REQUEST` or `CMD_LOAD_REQUEST` to send to SPM. Then we can entirely write our logic in controller but with higher latency. I think maybe the former method can have better performance with minimal addition to `DataMemController`.

cgra/CgraDmaRTL.py
HobbitQia: > All these will change if we go with @HobbitQia's option 2, right?
> 
> Moreover, can we use send/recv interfaces, and define msg (https://github.com/tancheng/VectorCGRA/blob/master/lib/messages.py) to encapsulate data, addr, or whatever needed as struct? So we don't need to declare so many ports, and explicitly connect each of them. This CGRA RTL shouldn't see these details, the data struct can be decomposed inside the submodule.

Yes.

I agree with we can use this helper functions or classes to define these ports. Initially I design the interface only from the side about how to connect the DMA engine and the chipyard so I defined these ports. But I believe these input/output ports can be used in our struct and wrapped by another adapter so that we can connect them. @BenkangPeng

tancheng:
> > Hi @HobbitQia, option 2 looks good to me. Though I am not sure what logic should be additionally in `DataMemController`?
> 
> I am thinking if we want enable the concurrent running of DMA and traditional load/store, then we need to multiplex the port of Data SPM and I think this logic can be implemented in `DataMemController`. However, we can also transform the data from DRAM into packets and use the command like `CMD_STORE_REQUEST` or `CMD_LOAD_REQUEST` to send to SPM. Then we can entirely write our logic in controller but with higher latency. I think maybe the former method can have better performance with minimal addition to `DataMemController`.

I thought they are the same latency if we can distinguish the `CMD_STORE_REQUEST` into `CMD_STORE_REQUEST_FROM_NOC` and `CMD_STORE_REQUEST_FROM_CPU` (and add another inport on the xbar) in controller?

Adding logic inside the `DataMemController` kind of bypassing the CGRA controller, which doesn't align with your Option 2, WDYT?

HobbitQia:
> > > Hi @HobbitQia, option 2 looks good to me. Though I am not sure what logic should be additionally in `DataMemController`?
> > 
> > 
> > I am thinking if we want enable the concurrent running of DMA and traditional load/store, then we need to multiplex the port of Data SPM and I think this logic can be implemented in `DataMemController`. However, we can also transform the data from DRAM into packets and use the command like `CMD_STORE_REQUEST` or `CMD_LOAD_REQUEST` to send to SPM. Then we can entirely write our logic in controller but with higher latency. I think maybe the former method can have better performance with minimal addition to `DataMemController`.
> 
> I thought they are the same latency if we can distinguish the `CMD_STORE_REQUEST` into `CMD_STORE_REQUEST_FROM_NOC` and `CMD_STORE_REQUEST_FROM_CPU` (and add another inport on the xbar) in controller?
> 
> Adding logic inside the `DataMemController` kind of bypassing the CGRA controller, which doesn't align with your Option 2, WDYT?

If the DMA data should go through the controller packet path, there may be extra latency of packeting, and there may be competitions between NoC/CPU/tile request to SPM? Or we have two separate paths in Controller?

tancheng:
> there may be extra latency of packeting

Packing can just be combinational logic before putting into a queue.

> and there may be competitions between NoC/CPU/tile request to SPM? Or we have two separate paths in Controller?

I was thinking about separate paths, so mentioned `FROM_CPU` and `FROM_NOC`. I feel the requests targeting same SPM bank should anyway conflict with each other, WDYT?

HobbitQia:
> > there may be extra latency of packeting
> 
> Packing can just be combinational logic before putting into a queue.
> 
> > and there may be competitions between NoC/CPU/tile request to SPM? Or we have two separate paths in Controller?
> 
> I was thinking about separate paths, so mentioned `FROM_CPU` and `FROM_NOC`. I feel the requests targeting same SPM bank should anyway conflict with each other, WDYT?

Got it. I mean the conflict between different paths. Even we have `FROM_CPU` and `FROM_NOC`, there may be concurrent read/write on SPMs. So maybe we need to multiplex the in/out ports of SPMs and the logic of multiplexing and handling the conflicts can be implemented in `DataSPMController`?

tancheng:
> > > there may be extra latency of packeting
> > 
> > 
> > Packing can just be combinational logic before putting into a queue.
> > > and there may be competitions between NoC/CPU/tile request to SPM? Or we have two separate paths in Controller?
> > 
> > 
> > I was thinking about separate paths, so mentioned `FROM_CPU` and `FROM_NOC`. I feel the requests targeting same SPM bank should anyway conflict with each other, WDYT?
> 
> Got it. I mean the conflict between different paths. Even we have `FROM_CPU` and `FROM_NOC`, there may be concurrent read/write on SPMs. So maybe we need to multiplex the in/out ports of SPMs and the logic of multiplexing and handling the conflicts can be implemented in `DataSPMController`?

Oh, we don't need to distinguish `FROM_CPU` and `FROM_NOC`, we can decompose the requests from CPU like: https://github.com/tancheng/VectorCGRA/blob/eb718426b7bd5b0aacad6761bc51c9b00d10b58c/controller/ControllerRTL.py#L207

HobbitQia:
> > > > there may be extra latency of packeting
> > > 
> > > 
> > > Packing can just be combinational logic before putting into a queue.
> > > > and there may be competitions between NoC/CPU/tile request to SPM? Or we have two separate paths in Controller?
> > > 
> > > 
> > > I was thinking about separate paths, so mentioned `FROM_CPU` and `FROM_NOC`. I feel the requests targeting same SPM bank should anyway conflict with each other, WDYT?
> > 
> > 
> > Got it. I mean the conflict between different paths. Even we have `FROM_CPU` and `FROM_NOC`, there may be concurrent read/write on SPMs. So maybe we need to multiplex the in/out ports of SPMs and the logic of multiplexing and handling the conflicts can be implemented in `DataSPMController`?
> 
> Oh, we don't need to distinguish `FROM_CPU` and `FROM_NOC`, we can decompose the requests from CPU like:
> 
> https://github.com/tancheng/VectorCGRA/blob/eb718426b7bd5b0aacad6761bc51c9b00d10b58c/controller/ControllerRTL.py#L207

But if we add DMA, there should be another path `FROM_DMA`, which is different from `FROM_CPU`/`FROM_NOC`?

tancheng:
> But if we add DMA, there should be another path FROM_DMA, which is different from FROM_CPU/FROM_NOC?

Ah, yes, makes sense. That `FROM_DMA` would be similar to `FROM_CPU`. The DMA_Controller then can assemble the different ports into our struct and send/recv interface.

HobbitQia:
> DMA_Controller

So this `DMA_Controller` refers to `DataSPMController` or our DMA engine?

tancheng:
> So this DMA_Controller refers to DataSPMController or our DMA engine?

DMA engine in your figure, or @BenkangPeng's `DmaEngineRTL`.

tancheng:
@BenkangPeng will you update this PR accordingly?

BenkangPeng:
> @BenkangPeng will you update this PR accordingly?

Yes, sorry for the delay. I will update this PR as soon as possible.

cgra/CgraTemplateRTL.py
```diff
@@ -135,6 +152,42 @@ def construct(s, CgraPayloadType,
...
+    # forwards DMA SPM access to the data memory.
+    if has_dma_ports:
+      s.dma_cmd = SendIfcRTL(DmaCmdType)
+      s.dma_cmd_val       = s.dma_cmd.val
+      s.dma_cmd_rdy       = s.dma_cmd.rdy
+      s.dma_cmd_opcode    = s.dma_cmd.msg.opcode
+      s.dma_cmd_dram_addr = s.dma_cmd.msg.dram_addr
+      s.dma_cmd_spm_addr  = s.dma_cmd.msg.spm_addr
+      s.dma_cmd_bytes     = s.dma_cmd.msg.nbytes
+      s.dma_cmd_tag       = s.dma_cmd.msg.tag
```
tancheng: Why do we decompose the `s.dma_cmd` into these sub-fields here? I didn't see any logic utilizing them. Even when we want to use them, they can be accessed as `s.dma_cmd.xxx`.

cgra/CgraTemplateRTL.py
```diff
@@ -135,6 +152,42 @@ def construct(s, CgraPayloadType,
...
+
+      s.dma_done = RecvIfcRTL(DmaDoneType)
+      s.dma_done_val      = s.dma_done.val
+      s.dma_done_rdy      = s.dma_done.rdy
+      s.dma_done_tag      = s.dma_done.msg.tag
```
tancheng: ditto

controller/ControllerRTL.py
```diff
@@ -52,6 +56,24 @@ def construct(s,
     YType = mk_bits(max(clog2(multi_cgra_rows), 1))
     TileIdType = mk_bits(clog2(num_tiles + 1))
     ControllerXbarPktType = mk_controller_noc_xbar_pkt(InterCgraPktType)
+    DmaOpcodeType = mk_bits(3)
+    DmaDramAddrType = mk_bits(64)
+    DmaBytesType = mk_bits(32)
+    DmaTagType = mk_bits(8)
+    DmaDramAddrPartType = mk_bits(32)
+    DmaMaskType = mk_bits(max(1, DataPayloadType.nbits // CHAR_BIT))
```
tancheng: These should be defined outside of this module, e.g., defined in `CgraDmaRTL`, and passed in when defining `ControllerRTL` inside the  `CgraDmaRTL`.

controller/ControllerRTL.py
```diff
@@ -75,6 +97,118 @@ def construct(s,
...
+    if has_dma_ports:
+      # Controller-owned command path from CPU packets to the DMA engine.
+      s.dma_cmd = SendIfcRTL(DmaCmdType)
+      s.dma_cmd_val       = s.dma_cmd.val
+      s.dma_cmd_rdy       = s.dma_cmd.rdy
+      s.dma_cmd_opcode    = s.dma_cmd.msg.opcode
+      s.dma_cmd_dram_addr = s.dma_cmd.msg.dram_addr
+      s.dma_cmd_spm_addr  = s.dma_cmd.msg.spm_addr
+      s.dma_cmd_bytes     = s.dma_cmd.msg.nbytes
+      s.dma_cmd_tag       = s.dma_cmd.msg.tag
```
tancheng: again, do we really need to decompose the `s.dma_cmd` into these explicit wires?

cgra/CgraTemplateRTL.py
BenkangPeng: Yes, you're right. They are unnecessary ports.
Fixed.

cgra/CgraTemplateRTL.py
BenkangPeng: Fixed.

controller/ControllerRTL.py
BenkangPeng: Fixed.

controller/ControllerRTL.py
BenkangPeng: Removed.

controller/ControllerRTL.py
```diff
@@ -52,6 +58,22 @@ def construct(s,
     YType = mk_bits(max(clog2(multi_cgra_rows), 1))
     TileIdType = mk_bits(clog2(num_tiles + 1))
     ControllerXbarPktType = mk_controller_noc_xbar_pkt(InterCgraPktType)
+    DmaOpcodeType = DmaCmdType.get_field_type('opcode')
```
tancheng: Similar to above attr, we should use `kAttrOpcode` instead of `opcode` here?

controller/ControllerRTL.py
```diff
@@ -52,6 +58,22 @@ def construct(s,
...
+    DmaDramAddrType = DmaCmdType.get_field_type('dram_addr')
```
tancheng: ditto for `dram_addr`?

cgra/CgraTemplateRTL.py
```diff
@@ -190,6 +220,17 @@ def construct(s, CgraPayloadType,
     s.data_mem.address_lower //= s.address_lower
     s.data_mem.address_upper //= s.address_upper
 
+    if has_dma_ports:
+      # CPU packets are decoded by the controller before becoming DMA commands.
+      s.dma_cmd  //= s.controller.dma_cmd
+      s.dma_done //= s.controller.dma_done
+
+      # DMA engine <-> controller side of the SPM path.
+      s.dma_spm //= s.controller.dma_spm_from_dma
```
tancheng: Rename this? `dma_spm` and `dma_spm_from_dma` sound confusing.

controller/ControllerRTL.py
BenkangPeng: Fixed.

controller/ControllerRTL.py
BenkangPeng: Fixed.

cgra/CgraTemplateRTL.py
BenkangPeng: How about `s.dma_spm_from_dma //= s.controller.dma_spm_from_dma`? My original idea was that `dma_spm` indicates the signal is used for communication between DMA and SPM, with suffixes like `from_dma` and `to_mem` indicating the direction of the signal.

cgra/CgraTemplateRTL.py
BenkangPeng: Fixed.

cgra/CgraDmaRTL.py
BenkangPeng: Fixed

cgra/CgraDmaRTL.py
```diff
@@ -0,0 +1,196 @@
...
+
+    s.cgra_id = InPort(CgraIdType)
+    # The local address range of current CGRA.
+    # Any address out of this range will be assumed as remote address.
+    s.address_lower = InPort(DataAddrType)
+    s.address_upper = InPort(DataAddrType)
+
+    # Abstract external dram memory interfaces for the internal DMA engine.
+
+    s.dram_rd_req = SendIfcRTL(DmaDramAddrType)
```
tancheng: Name this as `s.send_dram_rd_req`, to indicate its direction.

cgra/CgraDmaRTL.py
```diff
@@ -0,0 +1,196 @@
...
+    s.dram_rd_resp = RecvIfcRTL(DmaMemDataType)
```
tancheng: name this as `s.recv_dram_rd_resp`.

cgra/IntegratedCgraWithDmaRTL.py
```diff
@@ -0,0 +1,196 @@
...
+    s.dram_rd_resp = RecvIfcRTL(DmaMemDataType)
+
+    s.dram_wr_req = DmaDramWrReqIfcRTL(DmaDramAddrType, DmaMemDataType, DmaMemMaskType)
+
+    s.dram_wr_resp_val   = InPort()
+    s.dram_wr_resp_rdy   = OutPort()
+
+    # Components.
+
+    s.cgra = CgraTemplateRTL(CgraPayloadType,
```
tancheng, line 120: I was actually wondering why we need this `CgraDmaRTL.py`. Can we just expand the CgraTemplateRTL to include the DMA-related ports/interfaces?

lib/basic/val_rdy/ifcs.py
```diff
@@ -85,3 +85,99 @@ def construct( s, ReqType, RespType ):
...
+  """Undirected val/rdy stub interface for disabled DMA paths.
+
+  Uses OutPort for all signals so parent components can tie off unused
+  DMA stubs with structural connections during Verilog translation.
+  """
+  def construct( s, Type ):
+    s.msg = OutPort( Type )
+    s.val = OutPort()
+    s.rdy = OutPort()
+    s.trace_len = len(str(Type()))
```
tancheng: Can we just reuse the `ValRdySendIfcRTL`, but embed the `trace_len` into `msg`? The `msg` could have both `msg.data` and `msg.trace_len`. WDYT?

lib/basic/val_rdy/ifcs.py
```diff
@@ -85,3 +85,99 @@ def construct( s, ReqType, RespType ):
...
+    DMA-to-SPM Master Interface.
+    
+    This interface is instantiated on the DMA side. 
+    It initiates all transfer requests (both write and read) to the SPM 
+    and receives the corresponding read data back.
+    
+    Direction:
+    - write   : Output (Send). DMA sends write requests to SPM.
+    - read    : Output (Send). DMA sends read requests to SPM.
+    - read_resp: Input  (Recv). DMA receives read data from SPM.
```
tancheng: Can we reuse our send/recv interfaces, instead of creating new one?

lib/basic/val_rdy/ifcs.py
```diff
@@ -85,3 +85,99 @@ def construct( s, ReqType, RespType ):
...
+    DMA-to-SPM Minion Interface.
+    
+    This interface is instantiated on the SPM side.
+    It passively accepts incoming transfer requests from the DMA master, 
+    performs the requested memory operations, and returns read data if needed.
+    
+    Direction:
+    - write   : Input  (Recv). SPM receives write requests from DMA.
+    - read    : Input  (Recv). SPM receives read requests from DMA.
+    - read_resp: Output (Send). SPM sends read data back to DMA.
```
tancheng: ditto

lib/util/common.py
```diff
@@ -65,3 +65,28 @@
...
+StateType = mk_bits( 4 )
+STATE_IDLE          = StateType( 0 ) # Waiting for a new DMA command
+STATE_MVIN_REQ      = StateType( 1 ) # MVIN: Issuing DRAM read request
+STATE_MVIN_RESP     = StateType( 2 ) # MVIN: Waiting for DRAM read response
+STATE_MVIN_WRITE    = StateType( 3 ) # MVIN: Writing unpacked words to SPM
+STATE_MVOUT_READ    = StateType( 4 ) # MVOUT: Issuing SPM read request
+STATE_MVOUT_RESP    = StateType( 5 ) # MVOUT: Receiving SPM read response and packing
+STATE_MVOUT_WRITE   = StateType( 6 ) # MVOUT: Issuing DRAM write request
+STATE_MVOUT_WAIT    = StateType( 7 ) # MVOUT: Waiting for DRAM write response
+STATE_DONE          = StateType( 8 ) # Signaling command completion
```
tancheng: Change from `STATE_xxx` to `STATE_DMA_xxx`.

lib/messages.py
```diff
@@ -199,6 +199,130 @@ def str_func(s):
...
+  SpmDataType = mk_bits(spm_data_nbits)
+  SpmMaskType = mk_bits(spm_mask_nbits)
+  new_name = f"{prefix}_{dram_data_nbits}_{dram_mask_nbits}_{spm_data_nbits}"
+
+  def str_func(s):
+    return f"dma_data(dram_data={s.dram_data},dram_mask={s.dram_mask},spm_data={s.spm_data})"
+  
+  return mk_bitstruct(new_name, {
+    'dram_data': DramDataType,
+    'dram_mask': DramMaskType,
```
tancheng, line 280: explain what is `dram_mask` with comment?

lib/messages.py
```diff
@@ -199,6 +199,130 @@ def str_func(s):
...
+    'spm_data': SpmDataType,
+    'spm_mask': SpmMaskType,
```
tancheng, line 284: explain what is `spm_mask` with comment?

cgra/CgraTemplateRTL.py
tancheng: We should have two interfaces, one send and one recv, WDYT? and the "send"/"recv" should be in the name.

tancheng:
> Hi @tancheng @BenkangPeng , I summarized two direction of DMA design as below:
> 
> * Rely on data controller
>   
>   * DMA is added as a new client of the DataMemControllerRTL, where data in the DMA engine communicates directly with the DataMemControllerRTL, and the logic for multiplexing SPM ports is also implemented in that module.
>     To initiate DMA, the CPU can send `dma_mvin` or `dma_mvout` to the CGRA, after which the controller activates the DMA  engine by sending start signals.
>   * **Pros**: Keeps the controller clean; provides a faster path because data does not go through the controller.
>   * **Cons**: Additional logic is required to feed DMA results into the control memory.
>   
>      image
> * All in controller
>   
>   * All decoding logic is handled in the controller. The logic for handling the logic port should still reside in the DataMemControllerRTL, since the data memory should have its own port multiplexing logic.
>     The logic of packeting should also be implemented in the controller module.
>   * **Pros**: Unifies control and data memory within the controller (the controller is already connected to both control and data memory).
>   * **Cons**: Introduces complex control logic in the controller; results in a slower path.
>   
>      image
> 
> I prefer the second method but I think there are still some logic should be written in DataMemControllerRTL. WDTY?

We should also include the figure 2 into [our](https://github.com/tancheng/VectorCGRA/tree/master/doc/figures)

controller/ControllerRTL.py
```diff
@@ -358,6 +507,25 @@ def update_received_msg():
         #   # TODO: Handle other cmd types.
         #   assert(False)
 
+      if has_dma_ports & s.dma_done.val:
```
HobbitQia, line 510: I am thinking the possible conflict between CMD_DMA_CONE and CMD_COMPLETE or other commands that will be sent back to CPUs. Maybe we need some logic to judge these conflicts if at the same cycle the CMD_DMA_DONE and CMD_COMPLETE is valid?

mem/dma/DmaEngineRTL.py
```diff
@@ -0,0 +1,292 @@
...
+        s.beat_ff       <<= MemDataType( 0 )
+        s.word_idx_ff   <<= b2( 0 )
+        s.wr_mask_ff    <<= MemMaskType( 0 )
+      else:
+        if s.state == STATE_IDLE:
+          if s.dma_cmd.val & s.dma_cmd.rdy: # Receives a new DMA command.
+            s.opcode_ff     <<= s.dma_cmd.msg.opcode
+            s.dram_addr_ff  <<= s.dma_cmd.msg.dram_addr
+            s.spm_addr_ff   <<= s.dma_cmd.msg.spm_addr
+            s.words_left_ff <<= s.dma_cmd.msg.nbytes >> 2 # Converts the transfer size from bytes to words.
```
HobbitQia: There may be a truncation? If we want transfer 3 bytes, then `words_left_ff = 0` so no data will be transferred.

For now I think we don't need to consider such fine-grained data movement since usually DMA should handle the large-scale data. I think we can add some assertions or checking to ensure `nbytes%4 == 0` ?

mem/data/DataMemControllerRTL.py
```diff
@@ -223,6 +298,41 @@ def assemble_xbar_pkt():
...
+                                               s.cgra_id,                   # src_cgra
+                                               0,                           # src_tile
+                                               0)                           # remote_src_port
+
+        recv_waddr_from_dma = trunc(s.dma_spm.write.msg.addr, AddrType)
+        if (recv_waddr_from_dma >= s.address_lower) & (recv_waddr_from_dma <= s.address_upper):
+          bank_index_store_from_dma = trunc((recv_waddr_from_dma - s.address_lower) >> per_bank_addr_nbits, XbarOutWrType)
+        else:
+          bank_index_store_from_dma = XbarOutWrType(num_banks_per_cgra)
+        s.wr_pkt[dma_wr_idx] @= MemWritePktType(dma_wr_idx,                 # src
```
HobbitQia, line 294: Is this mask actually used?

cgra/IntegratedCgraWithDmaRTL.py
BenkangPeng: I initially designed the DMA module first. To integrate the CGRA with the DMA, I wrote a wrapper module(CgraDmaRTL) that instantiates and connects both the CGRA and the DMA.

I feel that directly instantiating and connecting a DMA inside CgraTemplateRTL would make the CGRA bloated. After all, many use cases don't require a DMA at al, and it doesn't align well with hierarchical modular design. I thought the CGRA and DMA should be separate modules, and a higher-level module should instantiate and connect them. WDYT?

tancheng: Makes sense. Then let's rename this file/module to `IntegratedDmaWithCgraRTL`?

And what about multi-cgra then? You will have another `IntegratedDmaWithMultiCgraRTL` that contains one DMA and a multi-cgra, correct?

tancheng:
> > Hi @tancheng @BenkangPeng , I summarized two direction of DMA design as below:
> > 
> > * Rely on data controller
> >   
> >   * DMA is added as a new client of the DataMemControllerRTL, where data in the DMA engine communicates directly with the DataMemControllerRTL, and the logic for multiplexing SPM ports is also implemented in that module.
> >     To initiate DMA, the CPU can send `dma_mvin` or `dma_mvout` to the CGRA, after which the controller activates the DMA  engine by sending start signals.
> >   * **Pros**: Keeps the controller clean; provides a faster path because data does not go through the controller.
> >   * **Cons**: Additional logic is required to feed DMA results into the control memory.
> >      image
> > * All in controller
> >   
> >   * All decoding logic is handled in the controller. The logic for handling the logic port should still reside in the DataMemControllerRTL, since the data memory should have its own port multiplexing logic.
> >     The logic of packeting should also be implemented in the controller module.
> >   * **Pros**: Unifies control and data memory within the controller (the controller is already connected to both control and data memory).
> >   * **Cons**: Introduces complex control logic in the controller; results in a slower path.
> >      image
> > 
> > I prefer the second method but I think there are still some logic should be written in DataMemControllerRTL. WDTY?
> 
> We should also include the figure 2 into [our](https://github.com/tancheng/VectorCGRA/tree/master/doc/figures)

@BenkangPeng, as you introduce `IntegratedDmaWithCgraRTL`, so we should update this diagram that the "Controller", "DataMemController", "SPM", "Control SPM" should be wrapped by a CGRA box. And entire diagram (except the "CPU") should be wrapped by "IntegratedDmaWithCgra" box.

cgra/IntegratedCgraWithDmaRTL.py
BenkangPeng: > And what about multi-cgra then? You will have another `IntegratedDmaWithMultiCgraRTL` that contains one DMA and a multi-cgra, correct?

Yes, I think so.

mem/dma/DmaEngineRTL.py
BenkangPeng: > There may be a truncation? If we want transfer 3 bytes, then `words_left_ff = 0` so no data will be transferred.
> 
> For now I think we don't need to consider such fine-grained data movement since usually DMA should handle the large-scale data. I think we can add some assertions or checking to ensure `nbytes%4 == 0` ?

Yes, I ignore it. I will add an assertion to ensure it.

BenkangPeng: Now the data transfer granularity between DRAM and SPM is 1 word (4 bytes).

* dram_mask: 16-bit, one bit per byte of 128-bit(16 bytes) DRAM beat.
* spm_mask: 4-bit, one bit per byte of 32-bit(4 bytes) SPM word.

HobbitQia:
I summarized the TODOs and something that needs to be clarified here as below. If there is something missing, please feel free to add it. @tancheng @BenkangPeng 

This PR should address the following items:

- Potential conflict between `CMD_DMA_DONE` and `CMD_COMPLETE`. (Marked as a future TODO)
- Double-check: synthesizability of constructs like `if has_dma_ports & s.dma_done.val:`, and correctness of the generated Verilog logic.
- Multi-CGRA support: a single DMA engine forwards data to the controller, which routes the data to the corresponding CGRA instances. (Marked as a future TODO)
- Illustrate the complete data paths:
    - Forward path: DRAM → DMA Engine → Controller → Data SPM Controller
    - Reverse path: Data SPM Controller → Controller → DMA Engine → DRAM
    
    Details to be clarified:
    
    - Data granularity (number of words transferred per cycle)
    - Transfer mode (serial vs. batch-parallel) and data packetization
    - Feasibility of extending the current implementation to support single-packet transfer instead of multi-cycle serial transfer
    - Internal FSM design
    - Update the architecture diagram with the above details

cgra/CgraTemplateRTL.py
```diff
@@ -135,6 +147,18 @@ def construct(s, CgraPayloadType,
...
+    # Optional DMA engine-facing ports. The controller owns command decode and
+    # forwards DMA SPM access to the data memory.
+    if has_dma_ports:
+      s.dma_cmd = SendIfcRTL(DmaCmdType)
+
+      s.dma_done = RecvIfcRTL(DmaDoneType)
+
+      s.dma_spm = DmaSpmMinionIfcRTL(DmaSpmWriteReqType,
+                                     DmaSpmReadReqType,
+                                     DmaSpmReadRespType)
```
tancheng: We shouldn't have this ifc as we discussed before? i.e., leverage existing SendIfcRTL/RecvIfcRTL.

cgra/IntegratedCgraWithDmaRTL.py
tancheng, line 1: Didn't we mention that we should rename this file to `IntegratedCgraWithDmaRTL.py`? Or sth like that?

cgra/IntegratedCgraWithDmaRTL.py
```diff
@@ -0,0 +1,196 @@
...
+
+    s.cgra_id = InPort(CgraIdType)
+    # The local address range of current CGRA.
+    # Any address out of this range will be assumed as remote address.
+    s.address_lower = InPort(DataAddrType)
+    s.address_upper = InPort(DataAddrType)
+
+    # Abstract external dram memory interfaces for the internal DMA engine.
+
+    s.send_dram_rd_req = SendIfcRTL(DmaDramAddrType)
```
tancheng: These ifcs are still not straightforward based on naming. I mean dram can be read, and can also issue rd request. Can we include direction into these interfaces? e.g., `send_rd_from_dram_req` or `send_rd_towards_dram_req`?

cgra/CgraTemplateRTL.py
BenkangPeng: yes, I will modify it. Some commits don't be push now since there are still some bugs to fix.

cgra/IntegratedCgraWithDmaRTL.py
BenkangPeng: How about `send_to_dram_rd_req` and `recv_from_dram_rd_resp`? They make it clear that the signal sends a read request to DRAM or receives the corresponding read response from dram.​

I've also seen similar naming conventions used elsewhere, e.g.:

https://github.com/tancheng/VectorCGRA/blob/eb718426b7bd5b0aacad6761bc51c9b00d10b58c/cgra/CgraTemplateRTL.py#L194-L199

WDYT?

cgra/CgraTemplateRTL.py
BenkangPeng: Fixed.

lib/basic/val_rdy/ifcs.py
BenkangPeng: Fixed.

lib/basic/val_rdy/ifcs.py
BenkangPeng: Fixed. I've removed all the new IfcRTL that I defined before and reused the current interfaces.

lib/util/common.py
BenkangPeng: Fixed.

lib/basic/val_rdy/ifcs.py
BenkangPeng: Fixed.

cgra/CgraDmaRTL.py
BenkangPeng: Is `s.recv_from_dram_rd_resp` okay?

cgra/CgraDmaRTL.py
BenkangPeng: I've named it as `s.send_to_dram_rd_req`.

cgra/CgraTemplateRTL.py
BenkangPeng: Fixed.

cgra/IntegratedCgraWithDmaRTL.py
BenkangPeng: Renamed.

lib/util/data_struct_attr.py
```diff
@@ -39,3 +39,14 @@
 kAttrDstCgraX = 'dst_cgra_x'
 kAttrDstCgraY = 'dst_cgra_y'
 kAttrAddr = 'addr'
+
+# DMA attributes
+kAttrOpcode = 'opcode'
+kAttrDramAddr = 'dram_addr'
+kAttrNBytes = 'nbytes'
+kAttrTag = 'tag'
```
tancheng: `tag` -> `dram_tag`, and comment on what this tag is used for?

lib/util/data_struct_attr.py
```diff
@@ -39,3 +39,14 @@
...
+kAttrSpmAddr = 'spm_addr'
+kAttrSpmData = 'spm_data'
```
tancheng, line 51: We already have `kAttrDataAddr`, but we are now distinguishing it from DRAM data. Can you file an issue, and do the cleanup (consolidate these attributes) in later PR?

cgra/IntegratedCgraWithDmaRTL.py
tancheng: Sounds good to me~!

mem/data/DataMemControllerRTL.py
```diff
@@ -120,7 +156,12 @@ def construct(s,
     s.send_to_noc_load_request_pkt = SendIfcRTL(NocPktType)
     s.send_to_noc_store_pkt = SendIfcRTL(NocPktType)
 
+    s.recv_from_ctrl_spm_wr_req = RecvIfcRTL(DmaSpmWriteReqType)
+    s.recv_from_ctrl_spm_rd_req = RecvIfcRTL(DmaSpmReadReqType)
+    s.send_to_ctrl_spm_rd_resp = SendIfcRTL(DmaSpmReadRespType)
```
tancheng: Why do we need to access `ctrl_spm` in `DataMemControllerRTL`?

tancheng (approved):

mem/data/DataMemControllerRTL.py
BenkangPeng: `recv_from_ctrl_spm_wr_req` means that receiving the request of writing into spm from `Controller`.

We connect the `Controller` with the `DataMemController`, and the `wr_req`, `rd_req`, `rd_resp` of SPM are transferred by these 3 ports.

I will update the current design diagram ASAP.

lib/util/data_struct_attr.py
BenkangPeng: `tag` -> `dma_tag`, right?

This tag isn't used now. Maybe we will use it to distinguish different DMA commands? @HobbitQia

lib/util/data_struct_attr.py
BenkangPeng: Oh, do you mean that we should rename `kAttrDataAddr`/`kAttrData` into `kAttrSpmAddr`/`kAttrSpmData`, since `kAttrDataAddr`/`kAttrData` of current codebase actually indicate the address/data of SPM? Is my understanding correct?

tancheng: Right, I feel some kAttr need to be refactored, but we don't need to do that in this PR.

lib/util/data_struct_attr.py
tancheng: right, `dma_tag` sounds good. You then can leave comment like:
```
This tag isn't used now. We may use it to distinguish different DMA commands.
```

mem/data/DataMemControllerRTL.py
tancheng: then plz rename it to `recv_from_controller_spm_wr_req`.

The current name sounds like: https://github.com/tancheng/VectorCGRA/blob/eb718426b7bd5b0aacad6761bc51c9b00d10b58c/controller/ControllerRTL.py#L344

BTW, how to send to ctrl_ring then? We not yet distinguish ctrl from data?

lib/util/data_struct_attr.py
BenkangPeng: File an issue: https://github.com/tancheng/VectorCGRA/issues/316

mem/data/DataMemControllerRTL.py
BenkangPeng: Fixed.

lib/util/data_struct_attr.py
BenkangPeng: Fixed.

mem/data/DataMemControllerRTL.py
BenkangPeng: > BTW, how to send to ctrl_ring then? We not yet distinguish ctrl from data?

Currently, DMA only transfers data between DRAM and SPM, and is not used for transferring ctrl_package.

mem/data/DataMemControllerRTL.py
BenkangPeng: No. The `spm_mask` is always `0xffff` When the DMA transfers data to the SPM.

Detailed explanation: https://github.com/tancheng/VectorCGRA/pull/293/changes#diff-9a6e2f65554f6195d1f3390e482af6fea499d7e92d1f2811baaf2f77c5d5c40eR236-R257

controller/ControllerRTL.py
BenkangPeng: Added relevant warning comments to flag the potential conflict issue. Left it as TODO.

lib/util/data_struct_attr.py
tancheng: Plz cite the issue around these code, i.e.,
```
# TODO: https://github.com/tancheng/VectorCGRA/issues/316 -- Consolidates attributes.
```

tancheng:
Thanks @BenkangPeng, plz fix the github action failure. @HobbitQia plz resolve your comments if you think @BenkangPeng fixed them.

lib/messages.py
BenkangPeng: Added.

lib/messages.py
BenkangPeng: Added.

lib/util/data_struct_attr.py
BenkangPeng: Added.

mem/dma/DmaEngineRTL.py
```diff
@@ -0,0 +1,307 @@
...
+            # Note: the nbytes % 4 check is omitted from the update block
+            # because PyMTL3's AST translator does not support assert
+            # statements. It is enforced in construct() instead.
+            s.opcode_ff     <<= s.dma_cmd.msg.opcode
+            s.dram_addr_ff  <<= s.dma_cmd.msg.dram_addr
+            s.spm_addr_ff   <<= s.dma_cmd.msg.spm_addr
+            # Converts the transfer size from bytes to words.
+            # NOTE We only support nbytes that are multiples of 4 now.
+            # If nbytes is not a multiple of 4, we will add 1 to the number of words to transfer.
+            s.words_left_ff <<= (s.dma_cmd.msg.nbytes >> 2) if (s.dma_cmd.msg.nbytes % 4 == 0) else (s.dma_cmd.msg.nbytes >> 2) + 1
```
HobbitQia: I am not sure whether there will be optimized by PyMTL or Verilog? Since `x % 4 == 0` can be easily written as `x[1:0] == 0`, and it will use shifter and comparator to check the condition rather than relying on divisions.

mem/dma/DmaEngineRTL.py
```diff
@@ -0,0 +1,307 @@
...
+            # Pack the response from SPM into a 128-bit beat by left-shifting.
+            if s.word_idx_reg == b2( 0 ): # 1st word
+              s.beat_ff <<= concat( s.beat_reg[spm_data_nbits:spm_data_nbits*4],
+                                    s.recv_from_spm_rd_resp.msg.data )
+            elif s.word_idx_reg == b2( 1 ):
+              s.beat_ff <<= concat( s.beat_reg[spm_data_nbits*2:spm_data_nbits*4],
+                                    s.recv_from_spm_rd_resp.msg.data,
+                                    s.beat_reg[0:spm_data_nbits] )
+            elif s.word_idx_reg == b2( 2 ):
+              s.beat_ff <<= concat( s.beat_reg[spm_data_nbits*3:spm_data_nbits*4],
```
HobbitQia: ditto. 

Can we write `*3` and `*4` as shift operators like `x<<1` and `(x<<1) + x`?

mem/dma/DmaEngineRTL.py
HobbitQia: And as you commented in line 228 of `message.py`: Currently, only nbytes that are multiples of 4 are supported. But here it seems that we still support nbytes is not mutiples of 4 and we just round up?

BenkangPeng: > I am not sure whether there will be optimized by PyMTL or Verilog? Since `x % 4 == 0` can be easily written as `x[1:0] == 0`, and it will use shifter and comparator to check the condition rather than relying on divisions.

Yes, `x[1:0] == 0` would be better. Thanks.

BenkangPeng: > But here it seems that we still support nbytes is not mutiples of 4 and we just round up?

Yes, If the data being moved is not an integer number of words, the assertion will report an error and this line will not be used. I will remove it.

mem/dma/DmaEngineRTL.py
BenkangPeng: Fixed.

mem/dma/DmaEngineRTL.py
BenkangPeng: Fixed.

controller/ControllerRTL.py
```diff
@@ -75,6 +95,28 @@ def construct(s,
...
+    s.dma_done = RecvIfcRTL(DmaDoneType)
+
+    # Receive the request of writing into SPM from the DMA.
+    s.recv_from_dma_spm_wr_req = RecvIfcRTL(DmaSpmWriteReqType)
+    # Receive the request of reading from SPM from the DMA.
+    s.recv_from_dma_spm_rd_req  = RecvIfcRTL(DmaSpmReadReqType)
+    # Send the response of reading from SPM to the DMA.
+    s.send_to_dma_spm_rd_resp   = SendIfcRTL(DmaSpmReadRespType)
+
+    # Data memory side of the same SPM access path.
```
tancheng: > # Data memory side of the same SPM access path.

what do you mean by "same SPM"? And should we change "Data memory side" to "SRAM data memory", it is a bit confusing now.

controller/ControllerRTL.py
```diff
@@ -75,6 +95,28 @@ def construct(s,
...
+    # Send the request of writing into SPM to the data_mem controller.
+    s.send_to_mem_spm_wr_req   = SendIfcRTL(DmaSpmWriteReqType)
```
tancheng: How is this different from above "send_to_mem_store_request"?

tancheng: IIRC, this is the isolated and newly introduced ifc to distinguish from legacy `inter-cgra store_req`. We need to refactor their names, e.g., 
`s.send_to_mem_store_request` -> `s.send_to_sram_store_request_from_noc`

`s.send_to_mem_spm_rd_req` -> `s.send_to_sram_store_request_from_dma`

How does this sound? Similar to the load and store_response. And leave comment that we prefer to separate the two as dma can perform burst data movement.

controller/ControllerRTL.py
BenkangPeng: > what do you mean by "same SPM"?

Oh, seems that it's meaningless and can be removed.
I just mean that these 2 groups of ports form the same SPM access path with the controller as the bridge, while the recv_from_dma_spm_* / send_to_dma_spm_* signals are the DMA-facing half, and send_to_mem_spm_* / recv_from_mem_spm_* are the SRAM data_memory-facing half.

https://github.com/tancheng/VectorCGRA/pull/293/changes#diff-2f580b4a845e2ec77388cdddf2e70b2fd8c19e6705933a7f796484f5331bbbd4R104-R117

> And should we change "Data memory side" to "SRAM data memory", it is a bit confusing now.

Yes, fixed.

controller/ControllerRTL.py
BenkangPeng: Fixed in https://github.com/tancheng/VectorCGRA/pull/293/commits/39c56fbb2d27f3e01291c0045280bec245b74c14

Is my understanding right?

tancheng: Yes. I hope my suggestion make sense though..

controller/ControllerRTL.py
```diff
@@ -71,9 +91,38 @@ def construct(s,
...
+
+    # Controller-owned command path from CPU packets to the DMA engine.
+    # Send the decoded DMA command to the DMA engine.
+    s.dma_cmd = SendIfcRTL(DmaCmdType)
+    # Receive the DMA done signal from the DMA engine.
+    s.dma_done = RecvIfcRTL(DmaDoneType)
+
+    # -------------------------------------------------------
+    # SPM (SRAM) access path from the DMA engine.
+    # The DMA and the inter-tile NoC (above) each have their own
```
tancheng: The noc is actually inter-cgra noc, rather than inter-tile.
`inter-tile NoC` -> `inter-cgra NoC`.

lib/messages.py
```diff
@@ -199,6 +199,185 @@ def str_func(s):
...
+#
+# === Mask Design ===
+# Data transfer granularity between DRAM and SPM is 1 word (4 bytes)
+# The `dram_mask` and `spm_mask` fields define the bitwidth of byte
+# masks for DRAM and SPM data respectively.
+#
+# Actual mask *values* are generated independently by the DMA engine
+# FSM (see DmaEngineRTL), NOT carried in this struct:
+#
+# - dram_mask (16-bit, one bit per byte of 128-bit(16 bytes) DRAM beat):
```
tancheng: space between `128-bit` and `(16 bytes)`. And plz also mention one word is 4-byte?

lib/messages.py
```diff
@@ -199,6 +199,185 @@ def str_func(s):
...
+#   Dynamically computed during MVOUT (SPM -> DRAM) based on the
+#   number of valid words in the last beat. Values range from 0x000f
+#   (1 word) to 0xffff (full beat). For example, if DMA move 1 word from SPM to DRAM, the mask is 0x000f.
+#   If DMA move 2 words from SPM to DRAM, the mask is 0x00ff.
+#   If DMA move 3 words from SPM to DRAM, the mask is 0x0fff.
+#   If DMA move 4 words from SPM to DRAM, the mask is 0xffff.
```
tancheng, line 257: So for now, we always use 1 word or 4 words? Plz clarify here. (i thought we can only read 32-bit (33 actually but ignore the last predicate bit, right?), so we always use 1 word?

lib/messages.py
```diff
@@ -199,6 +199,185 @@ def str_func(s):
...
+#
+# - spm_mask (4-bit, one bit per byte of 32-bit SPM word):
+#   SPM writes always write full words, so the mask is
```
tancheng: `full words` -> `one full word`?

lib/messages.py
BenkangPeng: The SPM has a data width of 32 bits, while the DRAM has a data width of 128 bits. Therefore, within one clock cycle, only 1 word can be read from or written to the SPM, whereas 4 words can be read from or written to the DRAM.

lib/messages.py
BenkangPeng: Fixed.

lib/messages.py
BenkangPeng: Fixed.

lib/messages.py
tancheng: Then can you provide some comments around line 265 of mem/dma/DmaEngineRTL.py? Iguess those code are for sequentially sending DRAM data to SRAM depending on how many words we read from DRAM.

BenkangPeng: Added.

BenkangPeng: > Then can you provide some comments around line 265 of mem/dma/DmaEngineRTL.py? Iguess those code are for sequentially sending DRAM data to SRAM depending on how many words we read from DRAM.

This section of code implements the logic for calculating the mask used when writing data to DRAM.
https://github.com/tancheng/VectorCGRA/pull/293/changes#diff-ffa8279d783d1ec8a0ae5d407d28a5325a105cca2d3ea57361b6b7183656d1b6R265

tancheng (approved):

tancheng:
@BenkangPeng feel free to merge. Thanks, awesome work!

BenkangPeng:
> @BenkangPeng feel free to merge. Thanks, awesome work!

Thanks for your thorough review and patience~

PR #301, opened 2026-06-20
[feature] Add Im2colRTL data mover and unit test

zsjiang99:
Single-channel scalar 32-bit im2col engine that streams an HxW image from one memory port and writes the (kH*kW)x(Hout*Wout) lowered matrix to a second port. Unit test verifies three configurations (4x4/2x2/s1, 4x4/2x2/s2, 5x5/3x3/s1) against a software golden in both pure-PyMTL3 simulation and full --test-verilog translation.


tancheng (commented):
Can we have a new module to connect this Im2col to the systolic array, so that we can verify e2e functionality from image towards matmul?

zsjiang99:
Yes, just saw this message. I will do it on this weekend.

cgra/test/Im2colCgraE2E_test.py
tancheng, line 1: Instead of constructing im2col with CGRA as a harness inside this test, we would like to see a `integratedIm2ColWithCgraRTL.py`, and the Im2Col is connected to the CGRA via the controller. Can you draw a diagram about the architecture design?

zsjiang99: Hi Cheng, I just updated the content based on your suggestion. You can see the architecute design diagram in `ntegratedIm2ColWithCgraRTL.py`. The `Im2colEngineRTL` and `CgraRTL` are integrated into the same RTL module. The dataflow proceeds as follows:

1. The external `start_im2col` signal is connected to `engine.start`. Once triggered, `Im2colEngineRTL` runs first and converts the input image into a lowered matrix.
2. The engine then generates a sequence of `CMD_STORE_REQUEST` packets. These packets are sent into the CGRA first to preload the required activation data.
3. The `engine.done` signal is connected to the external `im2col_done` output. Once the engine finishes, the external CPU packet stream is allowed to enter the CGRA and continue with the normal configuration, launch, and query packets.

tancheng: Instead of inport and outport of `start` and `done`, can we make them as val/rdy interfaces and passing `CMD`?

And instead of text diagram, can you provide a png like: https://github.com/tancheng/VectorCGRA/pull/293#issuecomment-4623403865

zsjiang99: Sure, I just updated it using val/rdy interfaces. And here is the png diagram and description: 

Image

When the CPU sends a launch command to the Im2colEngineRTL, the engine starts processing the preloaded input image. The engine converts the image data into a lowered column-format activation matrix and emits a sequence of preload packets to the CGRA. After all packets have been sent, the engine asserts the engine.done signal. This signal is used by the arbiter to determine which packet stream should access the CGRA controller input. Before engine.done is asserted, the arbiter gives priority to the Im2col engine’s preload packet stream. Once engine.done is asserted, the arbiter switches ownership to the external CPU packet stream, allowing the CPU to send the normal CGRA configuration, launch, and query packets.

tancheng: It looks cool~! However, can you make the controller be the first module connected to CPU? similar to https://private-user-images.githubusercontent.com/89443407/602918770-e74ec357-f9ef-47fb-942b-dd14d9ab79b0.png?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3ODI3MTExNjksIm5iZiI6MTc4MjcxMDg2OSwicGF0aCI6Ii84OTQ0MzQwNy82MDI5MTg3NzAtZTc0ZWMzNTctZjllZi00N2ZiLTk0MmItZGQxNGQ5YWI3OWIwLnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNjA2MjklMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjYwNjI5VDA1Mjc0OVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWQ1OTM1MGI1YmY3MDhmMTRhZmE1MjJhYmZiMDQxMWQ2ZDQ4Y2E4M2Y3ZDMxMmQ2MDZkZGNkZmYwOWRmNWExYTYmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JnJlc3BvbnNlLWNvbnRlbnQtdHlwZT1pbWFnZSUyRnBuZyJ9.yHhI44FI6cMUnJM8o9rHyiPCcfHgaQ7vZolrjQtY1_w

Your Img2Col should look like the DMA engine in above diagram.

zsjiang99: Yeah, but I can't open this website and find the figure. Could you provide another website or figure as a reference.

tancheng: Screenshot 2026-06-28 at 11 47 06 PM

Sorry, how about now? Basically, I want the controller be a universal adapter/interface for different modules/IPs.

zsjiang99: Yeah, got it. I will do it.

zsjiang99: Hi Cheng, I just modified this module. I designed a new CMD type and updated the controller so it can act as an adapter and transfer this CMD to the IM2COL engine.
image

tancheng:
Plz resolve the conflicts.

@BenkangPeng plz help review the PR.

zsjiang99:
Hi, Cheng. I just updated the `Controller.py`and `cmd_type.py` and resolve the conflicts in both files.

fu/single/Im2colRTL.py
```diff
@@ -0,0 +1,165 @@
...
+              s.kx    <<= AddrType(0)
+              s.ky    <<= AddrType(0)
+              s.ox    <<= AddrType(0)
+              s.oy    <<= s.oy + AddrType(1)
+              s.state <<= S_READ
+            else:
+              s.state <<= S_DONE
+
+        elif s.state == S_DONE:
+          if s.start:
```
BenkangPeng: Hi @zsjiang99 , seems that this `if` is unnecessary because when `s.state` equals `S_DONE(3)`, it's definitely not 0, right?

cgra/IntegratedIm2ColWithCgraRTL.py
```diff
@@ -0,0 +1,204 @@
...
+          s.engine_active <<= b1(0)
+
+    @update
+    def preload_pkt_arbiter():
+      s.engine.send_pkt.rdy   @= b1(0)
+      s.recv_from_cpu_pkt.rdy @= b1(0)
+
+      if s.engine_active:
+        s.cgra.recv_from_cpu_pkt.val @= s.engine.send_pkt.val
+        s.cgra.recv_from_cpu_pkt.msg @= s.engine.send_pkt.msg
```
BenkangPeng: Hi @zsjiang99 , why does the `img2colEngine` send the package to cgra here? As I know, `cgra.recv_from_cpu_pkt` is used to receive the config/ctrl packages from the CPU, right?

fu/single/Im2colRTL.py
zsjiang99: Hi @BenkangPeng, thanks for your review. You are right that `s.state` is not zero, but the variable in the if statement is `s.start`, not `s.state`. I think it is necessary to check the value of `s.start`. That means that in the DONE state, the module only returns to IDLE when the external side asserts start again; otherwise, it stays in DONE.

cgra/IntegratedIm2ColWithCgraRTL.py
zsjiang99: Yes. In the current design `IntegratedIm2ColWithCgraRTL.py`, we want the controller to serve as a universal adapter/interface for different modules/IPs. Although the port is named recv_from_cpu_pkt, it is actually a general packet ingress and is not limited to traffic originating from the CPU; it also carries Im2col-related traffic.

The CPU first sends an Im2col launch packet to the CGRA controller. The controller forwards the Im2col launch command to the compute engine, and the engine then sends its preload packets back through the same controller-facing ingress. Once `engine.done` is asserted, the arbiter returns control of this port to the external CPU.

tancheng: > "and the engine then sends its preload packets"

What are "preload" packets? Are they the data generated by the Im2Col? Can we create standalone new interfaces for those data, rather than leveraging the `recv_from_cpu_pkt`?

zsjiang99: The “preload packets” are the packets generated by the Im2col engine based on the lowered activations, not the raw image data. Sure, I will rename the port and update the comment to make this clearer.

tancheng: I was talking about the generated data after consuming the raw data. I meant to distinguish the `recv_from_cpu_pkt` from the `recv_from_img2col_pkt`. WDYT?

CC @BenkangPeng.

zsjiang99: Yes, LGTM. I just updated `ControllerRTL.py` and `IntegratedIm2ColWithCgraRTL.py`, and added `recv_from_img2col_pkt`.

cgra/Im2colEngineRTL.py
```diff
@@ -0,0 +1,217 @@
...
+    s.emit_idx = Wire(IdxType)
+
+    # Combinational dst/addr lookup (unrolled chain over emit_idx).
+    s.cur_dst  = Wire(TileIdType)
+    s.cur_addr = Wire(DataAddrType)
+
+    # Per-output constant ROMs as Wire arrays. The verilator translator
+    # accepts indexed access into Wire-arrays of bitstructs natively;
+    # plain Python lists / tuples of pymtl3-typed values do not translate.
+    s.dst_tile_rom  = [Wire(TileIdType)   for _ in range(num_outputs)]
```
BenkangPeng: Hi @zsjiang99 
The `_rom` suffix on `dst_tile_rom` / `data_addr_rom` is a bit misleading, since these are purely combinational wire arrays holding elaborated constants, not sequential ROM structures with address. Maybe we can consider renaming to `_lut` or `_const` to better reflect what they are. WDYT?

cgra/test/Im2colCgraE2E_test.py
```diff
@@ -0,0 +1,448 @@
+"""
```
BenkangPeng, line 1: It would be better to add a `generate_verilog` test to check whether the `IntegratedIm2ColWithCgra` is synthesizable.

like https://github.com/tancheng/VectorCGRA/blob/9fa363ae35571b729a5dbf85e95d93b1fd919e4c/cgra/test/IntegratedCgraWithDmaRTL_test.py#L271

Or just add a `pytest xxx_test.py --test-verilog --dump-vtb --dump-vcd` in CI yaml file, like

https://github.com/tancheng/VectorCGRA/blob/9fa363ae35571b729a5dbf85e95d93b1fd919e4c/.github/workflows/python-package.yml#L97

BenkangPeng (approved):
Looks good to me!

cgra/Im2colEngineRTL.py
zsjiang99: Yes, I think it will be clearer. I have changed their names to `dst_tile_const` and `data_addr_const`. Thanks for your review and suggestions, @BenkangPeng !

cgra/test/Im2colCgraE2E_test.py
zsjiang99: Sure, I just added it.

controller/ControllerRTL.py
```diff
@@ -249,6 +287,7 @@ def update_received_msg():
       kFromCpuCtrlAndDataIdx = 3
       kFromInterTileRingIdx = 4
       kFromReduceUnitIdx = 5
+      kFromIm2colIdx = 6
```
tancheng: Move these into the lib/util/common?

fu/single/Im2colRTL.py
tancheng, line 1: Instead of put it inside `fu/single`, can we move this into `fu/others`?

cgra/Im2colEngineRTL.py
tancheng, line 1: Move this file into `fu/others`, and explicitly mention its difference from the `Im2ColRTL.py`.

cgra/Im2colEngineRTL.py
```diff
@@ -0,0 +1,218 @@
...
+      s.data_addr_const[i] //= DataAddrType(data_addrs[i])
+
+    @update
+    def select_dst_and_addr():
+      s.cur_dst  @= s.dst_tile_const[0]
+      s.cur_addr @= s.data_addr_const[0]
+      for i in range(num_outputs):
+        if s.emit_idx == IdxType(i):
+          s.cur_dst  @= s.dst_tile_const[i]
+          s.cur_addr @= s.data_addr_const[i]
```
tancheng: so the `cur_addr` is the target spm on tile address to store the generated data, right?

it seems it is constant data provided by users?

cgra/Im2colEngineRTL.py
```diff
@@ -0,0 +1,218 @@
...
+    for i, v in enumerate(preload_image):
+      preload_full[in_base + i] = DataType(v, 1)
+
+    # Sub-components.
+    s.im2col  = Im2colRTL(DataType, scratch_mem_size)
+    s.in_mem  = DataMemRTL(DataType, scratch_mem_size,
+                           rd_ports = 1, wr_ports = 1,
+                           preload_data = preload_full)
+    s.out_mem = DataMemRTL(DataType, scratch_mem_size,
+                           rd_ports = 1, wr_ports = 1)
```
tancheng: I don't quite understand why we need in and out mem.. I thought data is moved in from either CPU or DRAM. We may need a in_mem to hold the img, but why we need a out_mem? Can the generated data just be sent out towards CGRA sram?

cgra/test/Im2colCgraE2E_test.py
tancheng: Can you add your arch diagram into `doc/figures` and make it similar to https://github.com/tancheng/VectorCGRA/blob/master/doc/figures/arch_of_IntegratedCgraWithDmaRTL.png?

controller/ControllerRTL.py
zsjiang99: Ok, got it.

fu/single/Im2colRTL.py
zsjiang99: Got it.

cgra/Im2colEngineRTL.py
zsjiang99: Yes, so It's like: Im2colRTL = Stand-alone im2col (image-to-column) data mover.
Im2colEngineRTL = Im2colRTL + scratchpads + packet emission FSM.  I think we can merge them into a `Im2colEngineRTL.py`. It will be clearer.  I will do that.

cgra/Im2colEngineRTL.py
zsjiang99: Yes, that's right. So the addr and tile id are provided by user in our design framework. You can find an example of the concrete values (`engine_dst_tiles = [6, 6, 3, 3]`, `engine_data_addrs = [0, 1, 2, 3]`) in our test for the 2x2 GEMM case, as follows:
 row 2 (top):   [tile 6] --> [tile 7] --> [tile 8]    
 row 1 (mid):   [tile 3] --> [tile 4] --> [tile 5]     
 row 0 (bot):   [tile 0]          [tile 1]     [tile 2]     
that means the input values are streamed from tiles 6 and 3, the weights are loaded into tiles 4, 5, 7, and 8, and use them to complete the matrix multiplication.

cgra/Im2colEngineRTL.py
zsjiang99: Yes, so I treated the CGRA as a pure systolic array before. If it were only a systolic array, I think we would need to design input and output memories. But CGRA already has internal memory in our project, so this memory is somewhat redundant, I will delete it. For the input memory, I believe it is necessary, because we need a buffer to hold the image data while Im2col is being executed.

zsjiang99: I will do these changes and provide architecture png in the next few days. Please let me know if you have any questions.

cgra/Im2colEngineRTL.py
tancheng: `engine_dst_tiles = [6, 6, 3, 3], engine_data_addrs = [0, 1, 2, 3]` is for activation tensor or weight tensor? I thought activation needs to be stored into sram, why tile_id is still provided...

cgra/Im2colEngineRTL.py
tancheng: SGTM

cgra/Im2colEngineRTL.py
zsjiang99: It's for activation tensor. I see, there is indeed no need for a tile ID, I will delete it. Thanks!

fu/others/Im2colEngineRTL.py
```diff
@@ -0,0 +1,254 @@
...
+      s.in_col  @= ScratchAddrType(s.ox * stride + s.kx)
+      s.in_addr @= ScratchAddrType(in_base + s.in_row * W + s.in_col)
+
+    # Per-output SRAM addresses as combinational Wire arrays. The
+    # verilator translator accepts indexed access into Wire-arrays of
+    # bitstructs natively; plain Python lists/tuples of pymtl3-typed
+    # values do not translate.
+    s.data_addr_const = [Wire(DataAddrType) for _ in range(num_outputs)]
+    for i in range(num_outputs):
+      s.data_addr_const[i] //= DataAddrType(data_addrs[i])
```
tancheng: Instead of user providing data_addrs[], can we just assume it would start from 0, and increment from 0?

zsjiang99: Ok, I will do it later.

tancheng: any update?

zsjiang99: Hi Cheng. I recently revised our previous XAI accelerator work, so could I update our im2col work after I finish revising it?

tancheng: Take your time, but I think this fix should be straightforward.

zsjiang99: Ok, thanks. I will check and fix it when I get time.

fu/others/Im2colEngineRTL.py
```diff
@@ -0,0 +1,237 @@
...
+The input scratchpad is kept: the caller preloads the image into it at
+construct time so the engine has a stable local buffer to walk during
+im2col (modeling the "image already resident in a nearby SPM" boundary
+condition).
+
+State machine: IDLE -> READ -> EMIT -> DONE
+  IDLE: wait for `start`.
+  READ: assert the read on in_mem at in_addr(ky, kx, oy, ox).
+  EMIT: hold the incoming data, drive send_pkt with
+        (data_addr = emit_idx, data). On send_pkt fire, advance the
```
tancheng, line 41: Instead of using `emit_idx` port, can we create a cmd like https://github.com/tancheng/VectorCGRA/blob/f44c97a2e497df35123fbc00b87ba6eca2afda8a/lib/cmd_type.py#L76 we can call it ` CMD_IM2COL_CONFIG_DEST_SPM_ADDR`, so this module can recognize this command and use the passed in value as the base address. How does this sound? @BenkangPeng

BenkangPeng: Sounds good to me :-)

tancheng: As discussed, this will be in another PR. Filed https://github.com/tancheng/VectorCGRA/issues/333 on you.

tancheng:
Hi @zsjiang99, nice work for this PR, i filed https://github.com/tancheng/VectorCGRA/issues/333 on you.

I am merging this one, but plz work on https://github.com/tancheng/VectorCGRA/issues/333 once you have bandwidth. Thanks!

zsjiang99:
Ok, Got it!

PR #323, opened 2026-07-12
Check-in some changes to run the kernels correctly

Jackcuii:


tancheng (commented):
Hi @Jackcuii, thanks for the fix, some fixes related to the fu make sense to me (i didn't walk through all of them yet).

I saw you also check in `trace_output`, `validation`, `visualizer`. I am wondering whether these can be a sub-module as a standalone repo?

Moreover, can you let your agent to create one PR for each `fu/single/xxxRTL.py` and other `xxxRTL.py` (e.g., tile, xbar, register), so it would be much easier to review. WDYT?

Jackcuii:
> Hi @Jackcuii, thanks for the fix, some fixes related to the fu make sense to me (i didn't walk through all of them yet).
> 
> I saw you also check in `trace_output`, `validation`, `visualizer`. I am wondering whether these can be a sub-module as a standalone repo?

Yep, I am going to clean them up later.

> Moreover, can you let your agent to create one PR for each `fu/single/xxxRTL.py` and other `xxxRTL.py` (e.g., tile, xbar, register), so it would be much easier to review. WDYT?

Of course！ working on it😃

Jackcuii:
Superseded by focused PRs: #325 (YAML generator), #326 (basic vector/FIR kernels), #327 (SpMV), #328 (GEMM/GEMV), and #329 (convolution/SAD). Existing focused RTL fixes remain in #298, #303-#305, #307, and #309-#312. The replacement PRs exclude visualizer and trace_output artifacts.

Jackcuii:
Added cleanup commit ce34b9b:

- removed all visualizer/ files
- removed all tracked trace_output/ files
- removed validation artifacts not directly referenced by tracked test/test_from_yaml code
- retained only script_generator.py and the eight newly added YAML files used by tracked tests

Python syntax checks pass. Pytest collection from the repository root is blocked by the project package layout (`attempted relative import beyond top-level package`), unrelated to the removed validation files.

Jackcuii:
## Detailed RTL-level change summary

This comment covers every `*RTL.py` file currently changed by this PR relative to the latest `master`. It describes both **what changed** and **why the change is needed**.

### 1. `cgra/CgraRTL.py`

**What changed**

- `line_trace()` now obtains the optional structured trace logger through `get_trace_logger()`.
- When logging is enabled, the complete CGRA state is recorded once when `line_trace()` is invoked.
- The existing human-readable line trace is unchanged.

**Purpose**

The structured log provides cycle-by-cycle machine-readable state for debugging long-running compiled kernels. It makes it possible to correlate tile/FU activity, control state, routing behavior, and memory activity without parsing the existing free-form line trace. The logger remains optional, so normal simulations do not need to enable it.

### 2. `fu/single/GrantRTL.py`

**What changed**

- Replaced the single global `already_grt_once` bit with one bit per control-memory slot (`ctrl_addr`).
- Added `cur_already_grt_once` to select the state associated with the currently executing control slot.
- Added support for `OPT_GRT_ONCE_CONST`, which takes its value from the constant queue instead of a regular FU input.
- Both `OPT_GRT_ONCE` and `OPT_GRT_ONCE_CONST` set the per-slot state only after a successful output valid/ready handshake.
- Reset and clear now reset every slot independently.

**Purpose**

A tile can contain more than one `GRANT_ONCE` operation at different control addresses. A single shared bit caused the first operation to suppress all later operations on that tile. Per-slot state makes each scheduled operation independent. `OPT_GRT_ONCE_CONST` is needed for compiled kernels in which a loop-invariant or entry-block value is supplied by the constant queue and must receive a true predicate exactly once.

### 3. `fu/single/MemUnitRTL.py`

**What changed**

- For `OPT_STR`, both the write-address valid signal and write-data valid signal are now gated by the predicates of the address and data inputs.
- The existing valid/ready data-consumption behavior is otherwise unchanged.

**Purpose**

During pipeline drain cycles, a token may still have `val=1` while carrying `predicate=0`. Previously, `OPT_STR` treated such a token as a real memory write and could overwrite a correct result with invalid drain data. Requiring both input predicates prevents side effects from predicated-off stores while still consuming the pipeline token normally.

### 4. `fu/single/PhiRTL.py`

**What changed**

- Added handling for `OPT_CONST` inside the Phi FU.
- The operation forwards the constant payload, combines the constant predicate with `reached_vector_factor`, and implements the normal constant/output valid-ready handshake.
- The existing per-control-slot PHI first-execution state from `master` is retained.

**Purpose**

The compiled schedule can map a plain constant-producing operation onto a tile configuration that uses the Phi-capable FU. Supporting `OPT_CONST` here avoids rejecting or stalling that schedule and preserves normal predicate/vector-factor behavior.

### 5. `fu/single/ShifterRTL.py`

**What changed**

- Added `OPT_LLS_CONST` for logical left shift by a constant amount.
- The data operand comes from the selected FU input and the shift amount comes from the constant queue.
- Output predicate, valid, and ready signals account for both operands and `reached_vector_factor`.

**Purpose**

Compiled address and index calculations frequently contain shifts by a compile-time constant. Supporting a constant shift avoids consuming a second routed FU input and allows the YAML generator to represent these operations directly.

### 6. `mem/const/ConstQueueDynamicRTL.py`

**What changed**

- Added a `consume_pending` state bit.
- A constant read handshake is remembered when `send_const.rdy` occurs before `ctrl_proceed`.
- The pending consumption retires and advances the read cursor on the next `ctrl_proceed` pulse.
- Reset/clear resets both the read cursor and pending state.

**Purpose**

The previous implementation advanced the queue only when `send_const.rdy` and `ctrl_proceed` were asserted in the same cycle. At iteration boundaries, another tile submodule can delay `ctrl_proceed`, causing an already-consumed constant to remain at the head of the queue. Remembering the pending consumption prevents constant-order corruption and repeated constants.

### 7. `mem/ctrl/CtrlMemDynamicRTL.py`

**What changed**

- Added `CMD_CONFIG_GEP_STRIDE` to the set of controller commands forwarded to the tile element.
- Added the same command to the set that is acknowledged/consumed from the controller packet queue.

**Purpose**

Two-dimensional GEP operations require runtime configuration of their stride. Without forwarding and consuming this command, the configuration packet can stall in control memory and the FU never receives the required stride value.

### 8. `mem/register_cluster/RegisterClusterRTL.py`

**What changed**

- Relaxed `recv_data_from_routing_crossbar[i].rdy` so routing tokens can be consumed when:
  - the current operation is `OPT_NAH`,
  - the FU does not select this lane,
  - the register bank supplies the FU input for this lane, or
  - the FU output path is ready.
- The previous, more conservative `master` condition is retained next to the implementation as a comment for comparison.
- Most of the apparent full-file diff is line-ending normalization; the ready condition above is the functional change.

**Purpose**

Routing-crossbar tokens should not remain blocked merely because the FU is not consuming that lane. In compiled schedules this produced cycle gaps and could backpressure the routing network even when the token was irrelevant to the current FU operation. The relaxed condition drains unused routing tokens while preserving normal FU consumption when the lane is selected.

### 9. `noc/CrossbarRTL.py`

**What changed**

- Prologued inputs are treated as satisfied for the all-input-valid gate, allowing non-prologued traffic to proceed.
- A prologued output no longer requires downstream ready because no token should be sent on that output during the prologue.
- Physical input data is masked from `recv_valid_vector` while the corresponding route is in its prologue period.
- `recv_required_vector` excludes prologued routes and uses set-only/OR-style accumulation so multicast outputs sharing one input cannot overwrite one another's requirement.
- `send_required_vector` excludes prologued outputs.
- Output `val` and input dequeue logic use `recv_valid_or_during_prologue_allowing_vector`, while the required vectors ensure that prologued channels are neither sent nor consumed.

**Purpose**

The earlier crossbar could see stale physical data on a prologued route, attempt to send it, and then block the entire all-or-nothing multicast transaction when the destination was full. It could also let a bypassed/prologued route suppress unrelated valid traffic. These changes make prologue bypass explicit on both the input and output sides, prevent accidental consumption, preserve multicast requirements, and eliminate the observed SpMV deadlock.

### 10. `tile/TileRTL.py`

**What changed**

- Added `CMD_CONFIG_GEP_STRIDE` to the controller commands routed from the tile's controller interface into `CtrlMemDynamicRTL`.
- Removed two trailing spaces; those are formatting-only changes.

**Purpose**

This is the tile-level half of GEP stride delivery. `TileRTL` must accept and route the command before `CtrlMemDynamicRTL` can forward it to the element. Without this condition, the stride configuration would never enter the tile control path.

### Note about `CompRTL.py`

`CompRTL.py` and `CompRTL_test.py` were previously modified by this branch, but they are no longer part of the current PR diff. After `master` merged its constant-comparison predicate semantics, the conflict was resolved by taking the `master` versions. Constant comparisons therefore follow `master` and require the data predicate, constant predicate, and `reached_vector_factor`.

Jackcuii:
Update: The translation works well. The signal_a[signal_b] can be translated.
YoSys do report some comb loop warnings, but they seem not to be a new-introduced problem.

Jackcuii:
Update: The reason of CI test fail is the invalid var that Verilog not support

tancheng:
> Update: The reason of CI test fail is the invalid var that Verilog not support

what is the `invalid var` you are referring to?

cgra/test/CgraRTL_test.py
tancheng, line 1: The trace-related changes in this file can be reverted.

cgra/CgraRTL.py
tancheng, line 1: the changes in this file can be reverted.

hmm, if you think this can help debugging, maybe we can also leave it there? as long as it can be conditionally invoked.

lib/trace_logger.py
tancheng, line 1: we can check this in if you want.

mem/register_cluster/RegisterClusterRTL.py
tancheng, line 1: Plz refactor this file, and show the diff appropriately.

noc/CrossbarRTL.py
tancheng, line 1: plz figure out whether this file is just for renaming or logic changes..

tancheng:
Hi @Jackcuii, can you resolve the comments if you have bandwidth? I would like to merge some register bank related fixes after yours: https://github.com/tancheng/VectorCGRA/pull/322

Jackcuii:
> Hi @Jackcuii, can you resolve the comments if you have bandwidth? I would like to merge some register bank related fixes after yours: https://github.com/tancheng/VectorCGRA/pull/322

Hi Sir Tan, I was working on it～ However I caught a fever yesterday 🥵. Hopefully I will be better and back tomorrow.

Jackcuii:
> > Update: The reason of CI test fail is the invalid var that Verilog not support
> 
> 
> 
> what is the `invalid var` you are referring to?

Oh sorry It is a typo, I mean char😂

tancheng:
> > Hi @Jackcuii, can you resolve the comments if you have bandwidth? I would like to merge some register bank related fixes after yours: #322
> 
> Hi Sir Tan, I was working on it～ However I caught a fever yesterday 🥵. Hopefully I will be better and back tomorrow.

Ah, take care, get better soon!

mem/register_cluster/RegisterClusterRTL.py
```diff
@@ -90,8 +90,21 @@ def update_msgs_signals():
...
 
-        s.recv_data_from_routing_crossbar[i].rdy @= ((s.inport_opt.write_reg_from[i] == PORT_ROUTING_CROSSBAR) \
-                & (s.inport_opt.operation == OPT_NAH)) | s.send_data_to_fu[i].rdy
+        # Consume routing-crossbar data whenever the FU does not need this
+        # lane, the register bank supplies the FU, or the FU is ready.
+        s.recv_data_from_routing_crossbar[i].rdy @= \
+            (s.inport_opt.operation == OPT_NAH) | \
+            (s.inport_opt.fu_in[i] == 0) | \
+            reg_towards_fu | \
+            s.send_data_to_fu[i].rdy
```
tancheng, line 107: @yyan7223 does this make sense to you?

noc/CrossbarRTL.py
tancheng, line 1: @yyan7223 can you plz take a look at this change?

mem/register_cluster/RegisterClusterRTL.py
```diff
@@ -90,8 +90,21 @@ def update_msgs_signals():
             (s.reg_bank[i].send_data.val & reg_towards_fu)
         s.reg_bank[i].send_data.rdy @= s.send_data_to_fu[i].rdy
 
-        s.recv_data_from_routing_crossbar[i].rdy @= ((s.inport_opt.write_reg_from[i] == PORT_ROUTING_CROSSBAR) \
-                & (s.inport_opt.operation == OPT_NAH)) | s.send_data_to_fu[i].rdy
+        # Consume routing-crossbar data whenever the FU does not need this
+        # lane, the register bank supplies the FU, or the FU is ready.
+        s.recv_data_from_routing_crossbar[i].rdy @= \
+            (s.inport_opt.operation == OPT_NAH) | \
+            (s.inport_opt.fu_in[i] == 0) | \
```
tancheng: do you think we can remove this line? the "i" in `fu_in[]` is not the "i" for the `routing_crossbar`...

The 'fu_in[]` can shuffle the chosen inports.

noc/CrossbarRTL.py
```diff
@@ -232,14 +243,18 @@ def update_in_dir_vector():
     def update_rdy_vector():
       s.send_rdy_vector @= 0
       for i in range(num_outports):
-        # The `outport_towards_local_base_id` indicates the number of outports that go to other tiles.
+        # The `num_inports` indicates the number of outports that go to other tiles.
```
tancheng: I don't think we need to change this comment of `outport_towards_local_base_id` -> `num_inports`, the original comment is correct.

noc/CrossbarRTL.py
```diff
@@ -232,14 +243,18 @@ def update_in_dir_vector():
...
         # Specifically, if the compute already done, we shouldn't care the ones
-        # (i.e., i >= outport_towards_local_base_id) go to the FU's inports. In other words, we skip
+        # (i.e., i >= num_inports) go to the FU's inports. In other words, we skip
```
tancheng: ditto

noc/CrossbarRTL.py
```diff
@@ -232,14 +243,18 @@ def update_in_dir_vector():
...
         # the rdy checking on the FU's inports (connecting from crossbar_outport) if
         # the compute is already completed.
         if (s.in_dir[i] > 0) & \
            (~s.compute_done | (i < outport_towards_local_base_id)):
-          s.send_rdy_vector[i] @= s.send_data[i].rdy
+          # When prologue is active for this output's input, don't
+          # require the downstream channel to be ready -- we won't be
+          # sending any data through it during prologue anyway.
+          s.send_rdy_vector[i] @= s.send_data[i].rdy | \
```
tancheng: then can we change `s.send_rdy_vector` to `s.send_rdy_or_during_prologue_vector`?

noc/CrossbarRTL.py
```diff
@@ -248,7 +263,17 @@ def update_valid_vector():
...
+          # When prologue is active for this output's input, treat the
+          # input as not-valid even if data is physically present in the
+          # channel.  This prevents the crossbar from asserting
+          # send_data.val (and therefore potentially blocking on a full
+          # downstream channel) during prologue steps.  Without this
+          # masking, stale data that happens to sit in a channel can
+          # trick the crossbar into attempting a send, and if the
+          # destination channel is full the all-or-nothing semantics
+          # cause a deadlock.
+          s.recv_valid_vector[i] @= s.recv_data_val[s.in_dir_local[i]] & \
```
tancheng: Can we then change this `recv_valid_vector` to `recv_valid_excluding_prologue_vector`

mem/register_cluster/RegisterClusterRTL.py
tancheng: Or change it to sth like:
```
  bank_used_by_fu = (s.inport_opt.fu_in[0] == (i + 1))
  for slot in range(1, num_reg_banks):
    bank_used_by_fu = bank_used_by_fu | (s.inport_opt.fu_in[slot] == (i + 1))

  ...

  s.recv_data_from_routing_crossbar[i].rdy @= \
      (s.inport_opt.operation == OPT_NAH) | \
      ~bank_used_by_fu | \
      reg_towards_fu | \
      s.send_data_to_fu[i].rdy
```
though I feel it is over-complicated.

tancheng:
Hi @Jackcuii, as long as you can resolve my comments, it should be okay to check in. Plz try to resolve them ASAP. Thanks!

Jackcuii:
> Hi @Jackcuii, as long as you can resolve my comments, it should be okay to check in. Plz try to resolve them ASAP. Thanks!

Okay~ I am checking in the logics the comments mentioned in the branch one by one to make sure all the yaml tests can still pass.  It is somehow black-boxed😂

tancheng:
> > Hi @Jackcuii, as long as you can resolve my comments, it should be okay to check in. Plz try to resolve them ASAP. Thanks!
> 
> Okay~ I am checking in the logics the comments mentioned in the branch one by one to make sure all the yaml tests can still pass. It is somehow black-boxed😂

I understand that. Can you then ask your agent to apply my suggestions to see whether it works?

Jackcuii:
Sadly some comments will cause some tests to fail, I am working on it

Jackcuii:
new TODO: merge #322 to this branch after folks' reviews

tancheng:
> new TODO: merge #322 to this branch after folks' reviews

I just merged it

tancheng:
@Jackcuii I remember you mentioned that you would update the tests, and double-check the last one finished or not?

So then we can review again and merge this?

Jackcuii:
> @Jackcuii I remember you mentioned that you would update the tests, and double-check the last one finished or not?
> 
> So then we can review again and merge this?

omg I just found I did not push all the commits :rofl: here they come.

Some logics have been changed to merge with main :melting_face:

mem/register_cluster/RegisterBankRTL.py
```diff
@@ -114,13 +129,31 @@ def update_token_status():
...
+      # Normal writes require an empty destination. FU feedback may use the
+      # pending slot when it reads and writes the same register; this breaks
+      # the output-ready/input-ready cycle of an in-place operation while
+      # preserving the old token until the ctrl step completes.
+      same_register_fu_feedback = \
+          (s.inport_opt.write_reg_from[reg_bank_id] == PORT_FU_CROSSBAR) & \
+          (s.inport_opt.write_reg_idx[reg_bank_id] == \
+           s.inport_opt.read_reg_idx[reg_bank_id]) & \
+          (read_towards != READ_TOWARDS_NOTHING)
+      if enable_token_discipline:
```
tancheng: What is this `enable_token_discipline`... I feel this is related to the skid buffer PR? https://github.com/tancheng/VectorCGRA/pull/330

Maybe we should merge above PR first, then hopefully we don't need to touch this file in this PR?

tancheng:
Hi @Jackcuii, #330 is merged, plz sync :-)

tancheng:
Hi @Jackcuii, any update :-) ?

Jackcuii:
> Hi @Jackcuii, any update :-) ?

Hi Sir Tan, the hacky logics are removed and tests are passed. But Yosys seems to fail this time ... checking

multi_cgra/test/sv_test/header.sv
tancheng, line 1: Why do we need to touch this folder?

lib/messages.py
```diff
@@ -174,6 +180,9 @@ def str_func(s):
   # 3: towards both FU and routing_xbar
   field_dict[kAttrReadRegTowards] = [RegFromType for _ in range(num_fu_inports)]
   field_dict[kAttrReadRegIdx] = [RegIdxType for _ in range(num_fu_inports)]
+  # One means the next scheduled access to this register needs a live version.
+  # A feedback write replaces it; a predicated-away write leaves it unchanged.
+  field_dict[kAttrReadRegRetain] = [mk_bits(1) for _ in range(num_fu_inports)]
```
tancheng, line 185: This new filed of `read_reg_retain` is suspicious... What is it used for? Can we avoid adding it?

multi_cgra/test/sv_test/header.sv
Jackcuii: Let me check it out, seems related to the Yosys check

lib/messages.py
tancheng: Any update on this @Jackcuii?

Jackcuii: Sorry Sir Tan I am moving to Swizterland these days... :melting_face: 

It seems to be necessary:

  > read_reg_retain marks whether the register version read by the current ctrl
  > step has another scheduled reader before the next write. The register bank
  > only sees the current ctrl word, so without this information it cannot
  > distinguish a non-final read from the final read. Clearing on every read
  > breaks values shared by multiple ctrl steps, while never clearing blocks the
  > next write. The script generator computes this bit from the cyclic per-
  > register read/write order; it is not kernel-specific. We could avoid
  > exposing the field only by adding control-memory lookahead/reference
  > counting in hardware or by rewriting the schedule, both of which are
  > substantially more complex.

:thinking: maybe need to dive into the register?

tancheng: I don't get what it wants to achieve. "Clearing on every read breaks values shared by multiple ctrl steps"...

in which scenario/kernel it needs this "fix"? We already and skid buffer in master and enable holding reg's consumption if current ctrl has other consumers..

tancheng: @Jackcuii Can we revert the register related change, and exclude the failed test? We can investigate the register related issue in a standalone PR. How does this sound?

Jackcuii: > @Jackcuii Can we revert the register related change, and exclude the failed test? We can investigate the register related issue in a standalone PR. How does this sound?

I think it makes sense🧐 (I haven't done the a/b test on it, let me find the real problem (maybe it is just introduce by wrong merge and actually does not break any test😢） By the way it seems I can't have the regular meeting with folks since it becomes 3 am in my time region🥲 What about we have 1:1 after I make some progress？ Thanks~

tancheng: sure, plz revise this PR to exclude that change, and then we can schedule a 1:1.

tancheng: Any update on split this PR :-)

Jackcuii: It breaks GEMM and GEMV, others are good
#326
(To make sure we still have a all-pass version for AE, this old branch is kept)
I somehow know the reason (bad habit x_x, did not record it carefully when I debugged a few month ago)

tancheng: well, I still see `_retain` in #336. For #326, my intention is you can check in all changes from this PR except the changes related to the `_retain`...

tancheng: Any update on https://github.com/tancheng/VectorCGRA/pull/336#pullrequestreview-5063207020?

Jackcuii: Hi Sir Tan I think the left  `_retain`s are leftover and not function at all. I just remove them :)

tancheng: Cool, thanks! Can you also sync that PR with master for submission?

tancheng: I think we can now close this PR but make a new one that only contains the register-related fix and the missing tests? (as the other PR was just merged)

PR #324, opened 2026-07-14
Add CLAUDE.md with development guidance for AI coding agents

tancheng:
## Summary
Adds a `CLAUDE.md` at the repo root — the standard file Claude Code (and other AI coding agents) read at session start to learn how a repository works.

The content is distilled from the practical lessons of recent development sessions (PRs #320 and #322), prioritizing knowledge that is *not* discoverable by reading a file or two:

- **Environment**: the custom `pymtl3.1@yo-struct-list-fix` fork requirement, submodules, verilator/pkg-config, and how to recognize the local verilator-version import failure as a toolchain issue rather than a code bug.
- **Testing**: that the default pytest run does **not** exercise Verilog translation (`--test-verilog` does), `pytest.ini`'s hidden tracebacks, which test tiers take seconds vs. ~20 minutes, and that `assert N < N` failures mean a kernel deadlocked.
- **RTLIR constraints**: no `in`/`not in` on hardware values, no tuple free variables, silent no-op bit-slice `<<=` in `@update_ff` (use set/clear masks), and the constant-wire pattern that works instead.
- **Architecture**: the ctrl-step protocol (`ctrl_proceed`, per-consumer done-latching), val/rdy conventions including FUs that snoop operands without handshakes, the comb-loop hazard through FU `recv_in.rdy`→`send_out.rdy`, register-cluster armed/token semantics, the three tile variants that must stay in sync, and how kernels are encoded in tests.

Docs-only change; no RTL or test modifications.


CLAUDE.md
```diff
@@ -0,0 +1,119 @@
+# CLAUDE.md
```
ShangkunLi: Would ' AGENT.md ' be better? Since we also use `codex` for coding.

CLAUDE.md
```diff
@@ -0,0 +1,119 @@
...
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
```
ShangkunLi: Same problem: not just for `cc`, `codex` or other agents may also use this.

CLAUDE.md
```diff
@@ -0,0 +1,119 @@
...
+  `CMD_CONFIG_TOTAL_CTRL_COUNT` bookkeeping, then compare sink outputs.
+  Changing component semantics usually means re-deriving expected sink
+  sequences cycle-by-cycle; sinks (`lib/basic/val_rdy/SinkRTL.py`) check order,
+  not cycle numbers, and stop accepting once their expected list is exhausted.
+- The command-type dispatch in `ControllerRTL` forwards any inter-CGRA command
+  without a dedicated handler to the ctrl ring (catch-all), so new `CMD_*`
+  types work by default; `CtrlMemDynamicRTL`'s accept-list is curated and must
+  be extended explicitly.
+
+## Repo conventions
```
ShangkunLi: Shall we also add the coding style here?
For example: `snake_case` for variables, and `CamelCase` for function/class declarations/definitions.

CLAUDE.md
tancheng: Good point — done, with one tweak: the cross-tool convention Codex/Cursor/etc. actually read is `AGENTS.md` (plural), so the content now lives there with a tool-neutral header. Claude Code auto-loads `CLAUDE.md` rather than `AGENTS.md`, so `CLAUDE.md` is kept as a one-line `@AGENTS.md` import — both tools read the same single source of truth with no duplication.

CLAUDE.md
tancheng: Addressed together with the rename: the header is now tool-neutral ("AI coding agents (Claude Code, Codex, and others)") and explains the CLAUDE.md import indirection.

CLAUDE.md
tancheng: Added — with one correction from checking the codebase: functions here are actually `snake_case` (`update_msg`, `comb_logic`, `mk_ctrl`), not CamelCase; CamelCase is used for classes and constructed type names (`RegisterBankRTL`, `DataType`). The new bullet documents: CamelCase classes/types, snake_case functions/methods/variables, UPPER_SNAKE_CASE constants, and the `kCamelCase` attribute-key family (`kAttrPayload`).

ShangkunLi (approved):

Issue #286, opened 2026-04-23
[P0] Overwrite and overuse the same register in xbar


tancheng:
`s.send_rdy_vector[i] @= s.send_rdy_towards_other_tiles or s.send_accepted_towards_other_tiles` from https://github.com/tancheng/VectorCGRA/issues/281#issuecomment-4238976584

yyan7223:
Should be discussed during tomorrow's meeting.

tancheng:
Hi @yyan7223, any progress on this issue? Thanks~!

yyan7223:
Sorry, I was busy in the past weeks, but I will try to fix this before May 18.

PR #330, opened 2026-07-15
Add write skid buffer for same-register read-modify-write

tancheng:
Follow-up to #322 (based on its branch; retarget to `master` after #322 merges). Removes #322's known limitation and addresses the register half of #281/#286.

## Problem
Under #322's token discipline, an operation that reads and writes the **same register within one ctrl step** deadlocked: the write was rejected while the operand's token was unconsumed, and the step could not complete until the write was delivered. This pattern is real — the mapper emits it (`NOT $0 -> $0, SOUTH` in the histogram kernel from #281), and it is the natural encoding of an accumulator.

## Design
Each bank gains a **one-entry write skid buffer** (one write port ⇒ at most one blocked write, so one entry per bank suffices; the cluster/tile interfaces are unchanged).

- **Acceptance**: `outport_wr_rdy = ~skid_valid` — a write is accepted whenever the skid is free. Pure registered state: the producer's `rdy` never combinationally depends on any consumer's readiness (deliberately *not* on `skid_commit`, which derives from `ctrl_proceed` and would close a loop through the FU's rdy chain), and a direct write can never collide with a commit on the single write port.
- **Routing**: an accepted write lands **directly** if that cannot disturb anything (target token-free and not read by the current step), or **at the reading step's boundary** (write-through on the `ctrl_proceed` pulse — the new token atomically replaces the consumed one; set wins over clear). Otherwise it **parks** in the skid and commits once its target is no longer read by an open step.
- **Stability invariant**: a register being read by an open step never changes value until the step completes. This matters beyond RMW: the previous behavior (direct write whenever token-free) allowed #281's partial-multicast corruption — a speculative register write landing while another fan-out leg is backpressured, then re-execution on the changed operand. The tile-level test below **caught exactly this** during development.
- Skid state is cleared on task switch along with the token bookkeeping.

With the write-through path, a same-register accumulator runs at **full rate** (one iteration per step, no bubble).

## Tests
- `test_reg_cluster_write_skid_buffer`: three back-to-back writes to one register with a stalled consumer flow through park→commit→park; all three delivered in order.
- `test_tile_same_register_accumulate`: single-ctrl-word `INC` accumulator through one register that also fans out to a stalled tile outport — the #281 shape end-to-end at tile level. **Deadlocks at max-cycles against the #322 base (verified as a negative control); completes with the skid.** The unarmed first read seeds the accumulator with the register default, so the outport observes 1, 2, 3.
- Cluster test harnesses now always emulate the tile's per-step done-tracking (`ctrl_proceed`), since commit timing is tied to step completion; historical expectations are preserved, with stalled-sink tests observing one leading unarmed read.
- Kernel regression sweep (everything that ever failed on #322's CI rounds: `CgraRTL_test` ×3, FIR terminate, streaming ×2, vector global reduce, migration ×3): **10/10 pass** locally.

## Relation to #286
The xbar half of #281 (`send_rdy_vector` ignoring `send_accepted`) is separate and remains open in #286 — this PR handles the register-file half (stability + RMW liveness).


mem/register_cluster/RegisterBankRTL.py
```diff
@@ -111,16 +152,46 @@ def update_token_status():
...
+      # A write is accepted whenever the skid buffer is free: it lands
+      # directly in the register file if that cannot disturb anything,
+      # and parks in the skid otherwise. Occupied skid -> not ready.
+      # Keeping this a pure function of registered state means the
+      # producer's rdy never combinationally depends on any consumer's
+      # readiness (in particular not on skid_commit, which derives from
+      # inport_ctrl_proceed and would otherwise close a loop through
+      # the FU's rdy chain), and it also guarantees a direct write can
+      # never collide with a skid commit on the single write port.
+      s.outport_wr_rdy @= ~s.skid_valid
```
yyan7223: How about letting `outport_wr_rdy @= inport_ctrl_proceed`?

https://github.com/tancheng/VectorCGRA/blob/3d4d0571664479b73d9d517d40fb68cbbce436b7/tile/TileRTL.py#L304

https://github.com/tancheng/VectorCGRA/blob/3d4d0571664479b73d9d517d40fb68cbbce436b7/tile/TileRTL.py#L296-L298

https://github.com/tancheng/VectorCGRA/blob/3d4d0571664479b73d9d517d40fb68cbbce436b7/tile/TileRTL.py#L314-L319

There is no combinational loop as three registers `xxx_done` are involved. 

Semantically, `inport_ctrl_proceed=1` at cycle N means the token will be consumed at cycle N. Then the `outport_wr_rdy=1` at cycle N will allow the new token (either from FU computation result or routing buffer) being directly and safely written to register bank at the rising edge of cycle N+1. During cycles when `inport_ctrl_proceed=0`, the new token is either combinationally maintained on FU's outport or kept in the routing buffer (ChannelRTL) and `outport_wr_rdy=0` also avoids the unsafe overwriting. Seems the skid buffer is no longer necessary here.

Taking `NOT $0 -> $0, SOUTH` as an example, assuming the routing buffer of south tile has no space, then we have `inport_ctrl_proceed=0`, so the computation result of NOT (new token) will not overwrite $0 at the rising edge of cycle N+1 because `outport_wr_rdy=0`. If south tile suddenly has buffer space at cycle 2N, we will have `inport_ctrl_proceed=1` and `outport_wr_rdy=1` at cycle 2N, then the new token that is combinationally kept on FU's outport during cycle N~2N can safely overwrite $0 at the rising edge of cycle 2N+1.

mem/register_cluster/RegisterBankRTL.py
```diff
@@ -90,16 +99,48 @@ def construct(s, DataType, CtrlType, reg_bank_id, num_registers = 4):
...
     # configured read (acting as a default-token source).
     s.armed = Wire(num_registers)
+    # Write skid buffer (one entry; the bank has one write port, so at
+    # most one write can be blocked at a time). A write that cannot land
+    # directly is accepted ("parked") here and committed into the
+    # register file once that cannot disturb an in-flight read. This is
+    # what allows an operation to read and write the same register
+    # within one ctrl step (e.g., an accumulator, or `NOT $0 -> $0,
+    # SOUTH` under backpressure — see issues #281/#286): the producer's
+    # write handshake completes immediately via the skid, so the step
```
yyan7223: The producer here is the south ChannelRTL, which has already been a buffer, right? Why we need an additional skid buffer here?

mem/register_cluster/RegisterBankRTL.py
```diff
@@ -90,16 +99,48 @@ def construct(s, DataType, CtrlType, reg_bank_id, num_registers = 4):
...
     s.read_towards_xbar = Wire(1)
     s.read_token_valid = Wire(1)
     s.read_armed = Wire(1)
+    # Token status of the configured write target and of the skid
+    # entry's target.
+    s.wr_target_token = Wire(1)
+    s.skid_target_token = Wire(1)
+    # The current ctrl step is reading the write target / the skid
+    # entry's target. A register being read must stay stable until the
+    # step completes (reads are level signals), so such writes park in
```
yyan7223: Seems that `target token` refers to the new token, and `token` refers to the unconsumed token, right?

And what is the difference between `s.wr_target_token` and `s.skid_target_token`, could you please ask Claude to elaborate more on this?

mem/register_cluster/RegisterBankRTL.py
```diff
@@ -111,16 +152,46 @@ def update_token_status():
...
-          s.outport_wr_rdy @= ~s.token_valid[r]
+          s.wr_target_token @= s.token_valid[r]
+        if s.skid_idx == r:
+          s.skid_target_token @= s.token_valid[r]
+
+      s.wr_target_read @= (s.read_towards_fu | s.read_towards_xbar) & \
+          (s.inport_opt.read_reg_idx[reg_bank_id] == \
+           s.inport_opt.write_reg_idx[reg_bank_id])
+      s.skid_target_read @= (s.read_towards_fu | s.read_towards_xbar) & \
+          (s.inport_opt.read_reg_idx[reg_bank_id] == s.skid_idx)
```
yyan7223, line 192: Looks that these two signals are read enable signals? Either reading from register or skid buffer.

mem/register_cluster/RegisterBankRTL.py
```diff
@@ -111,16 +152,46 @@ def update_token_status():
...
+      s.skid_target_token @= 0
       for r in range(num_registers):
         if s.inport_opt.read_reg_idx[reg_bank_id] == r:
           s.read_token_valid @= s.token_valid[r]
           s.read_armed @= s.armed[r]
         if s.inport_opt.write_reg_idx[reg_bank_id] == r:
-          s.outport_wr_rdy @= ~s.token_valid[r]
+          s.wr_target_token @= s.token_valid[r]
+        if s.skid_idx == r:
+          s.skid_target_token @= s.token_valid[r]
```
yyan7223: These two are write enable signals? Why they are assigned to token_valid?

yyan7223 (commented):
The overall skid buffer control logic looks fine. But I need more elaboration to 4 signals below to help me better understand the details:
- s.wr_target_token
- s.skid_target_token
- s.wr_target_read
- s.skid_target_read

mem/register_cluster/RegisterBankRTL.py
tancheng: > There is no combinational loop as three registers xxx_done are involved.

But there are also `recv_opt.rdy` or (i.e., `|`) with the `xxx_done`, leading to loop:
```
bank.outport_wr_rdy (= inport_ctrl_proceed)
  → cluster.recv_data_from_fu_crossbar[i].rdy        (RegisterClusterRTL L98-99)
  → fu_crossbar.send_data[写腿].rdy                   (结构连接)
  → crossbar.send_rdy_vector → all_send_accepted      (CrossbarRTL update_rdy_vector / update_signal)
  → fu_crossbar.recv_opt.rdy                          (CrossbarRTL L148)
  → tile.send_ctrl.rdy                                (TileRTL L296-298)
  → ctrl_proceed → 回到 bank.outport_wr_rdy           ←←← 闭环
```

mem/register_cluster/RegisterBankRTL.py
tancheng: No. I think the producer is from `FU -> fu_crossbar -> register_cluster`, there is no channel.

mem/register_cluster/RegisterBankRTL.py
tancheng: "target" modifies *register*, not token: `wr_target_holds_token` (renamed in the latest push for clarity) is `token_valid[write_reg_idx]` — whether the register this write targets still holds its **old, unconsumed** token. It never refers to the value being written. The two signals ask that same question for two different writes: the incoming write of this cycle (target = the current ctrl word's `write_reg_idx`) vs. the write parked in the skid (target = `skid_idx`, latched at park time, possibly by an earlier step — so the two targets can differ). The first steers the incoming write (direct vs park); the second times the parked entry's commit.

mem/register_cluster/RegisterBankRTL.py
tancheng: Not write enables — they're the loop-unrolled mux for indexing a vector with a signal: `wr_target_holds_token = token_valid[write_reg_idx]`. Pure status *lookups* feeding the write-routing decision; the actual write enables are `wr_en` (direct write) and `skid_commit` (skid drain), computed in `access_registers`.

mem/register_cluster/RegisterBankRTL.py
tancheng: Not read enables — they're hazard flags: "is the current ctrl step reading the register that {this write / the parked entry} targets?" (renamed to `wr_target_being_read` / `skid_target_being_read` in the latest push). They force parking and time the commit so an in-flight read is never disturbed. Also worth noting: nothing ever reads *from* the skid — reads always come from the register file, and if a register's next value is parked, readers see the old register value until the step completes. That's the stability invariant, by design.

tancheng:
@yyan7223 Elaboration on the 4 signals (they were renamed in a recent push, exactly to address this — mapping below):

| Signal (as reviewed) | Renamed to | What it is | Role in the control logic |
|---|---|---|---|
| `s.wr_target_token` | `s.wr_target_holds_token` | `token_valid[write_reg_idx]` — whether the register targeted by **this cycle's incoming write** still holds its **old, unconsumed** token. ("target" = the destination *register*; it never refers to the value being written.) | Steers the accepted write: target token-free **and** not being read → lands directly in the register file; otherwise → parks in the skid buffer. |
| `s.skid_target_token` | `s.skid_target_holds_token` | `token_valid[skid_idx]` — the same question, asked for the register targeted by the **write parked in the skid** (`skid_idx` was latched at park time, possibly during an earlier ctrl step, so the two targets can differ). | Times the parked entry's drain when its target is *not* being read: commit once that old token has been consumed. |
| `s.wr_target_read` | `s.wr_target_being_read` | Hazard flag: the **current ctrl step is reading** the incoming write's target register (`read_reg_idx == write_reg_idx`, with a read configured). | A register being read must stay stable until the step completes, so this forces the write to park — or to land exactly on the `ctrl_proceed` pulse (write-through at the step boundary, where the new token atomically replaces the consumed one). |
| `s.skid_target_read` | `s.skid_target_being_read` | Hazard flag: the current ctrl step is reading the **parked entry's** target register (`read_reg_idx == skid_idx`). | Gates the skid commit to the `ctrl_proceed` pulse, so the drain also lands exactly at the step boundary instead of disturbing the in-flight read. |

Two clarifications that tie these together:
- The `token_valid[r]` assignments in the for-loop are not enables — they are the loop-unrolled mux for indexing a vector with a signal (RTLIR-friendly form of `token_valid[write_reg_idx]`). The actual write enables are `wr_en` (direct) and `skid_commit` (drain).
- Nothing ever reads *from* the skid: reads always come from the register file. If a register's next value is parked, readers see the old register value until the step completes — that is the stability invariant that fixes #281's mid-step corruption.

Branch is also synced with `master` (post-#322 merge), so the current diff shows the new names.

tancheng:
Fixed the two CI failures (`CgraRTL_fir_2x2_loop_counter_test` and `CgraVerifAssert_test`, both `2x2_fir_*_return`) in 912bdbf — they were a real regression from this PR, not pre-existing.

**Root cause.** The first version set `outport_wr_rdy = ~skid_valid`, i.e. it accepted *any* write while the skid was free and parked whatever could not land directly. That silently changed flow control for kernels that never do same-register RMW: a write whose target still held an unconsumed token used to backpressure its producer, but was now accepted into the skid, letting the producer run ahead. The two FIR kernels then computed `0x413` instead of `0x8a7`.

**How it was isolated.**
1. Baseline: reverting `mem/register_cluster/` to `master` and re-running both tests → both pass, so the regression is mine.
2. Bisection: removing only the read-hazard parking rule while keeping `wr_rdy = ~skid_valid` → still fails, so the acceptance condition was the culprit, not the stability rule.

**Fix.** The skid is now used *only* for the case that has no other solution — a write whose target is being read by its own ctrl step (backpressuring it deadlocks; landing it mid-step corrupts the in-flight read):

```python
s.outport_wr_rdy @= (~s.wr_target_holds_token & ~s.wr_target_being_read) | \
                    (s.wr_target_being_read & ~s.skid_valid)
```

The first term is the pre-existing acceptance condition, so backpressure for every flow that already worked is bit-for-bit unchanged; only same-register RMW takes the skid path. All terms are still registered state or ctrl-word fields, so `rdy` still never combinationally depends on a consumer's readiness, and a direct write still cannot collide with a skid commit on the single write port. Docstring/comments updated to match the narrowed rule.

**Verification:** 23/23 locally — both previously-failing tests, the full FIR 2x2 set, `CgraRTL_test` ×3, streaming ×2, 4x4 FIR terminate, vector global reduce, the migration suite, and all register-cluster tests (the skid tests are RMW-shaped by construction, so they still exercise the skid path).

mem/register_cluster/RegisterBankRTL.py
yyan7223: Is it possible to break this combinational logic `crossbar.send_rdy_vector→ all_send_accepted`? For example, making `all_send_accepted` a register instead of a wire in current design. Then the combimational loop above may disappear and we may get rid of the skid buffer.

The reason why I insist on my opinion is that from the first principle, one writing request to the register cluster either comes from fu or from other tiles. So when current ctrl is not proceeded, the new token can either be held in fu's outport or in channels, instead of writing to a redundant skid buffer. People who investigate the RTL design of the register cluster may question the semantic of the skid buffer.

mem/register_cluster/RegisterBankRTL.py
yyan7223: Oh, sorry.
But in other cases, the register cluster may receive inputs from channel, right?

mem/register_cluster/RegisterBankRTL.py
yyan7223: Bohan's progress is more important,  we should merge above PR first if all tests pass.

tancheng: > the new token can either be held in fu's outport or in channels

I understand your concern. But the issue is there is no latch from FU outport to register. I introduced the register bank/cluster to serve as the latch originally.. We can discuss this fundamental design problem later.

mem/register_cluster/RegisterBankRTL.py
tancheng: right, any issue in other cases?

yyan7223 (approved):

Issue #333, opened 2026-07-23
[P1] Configure the base addr and input size for im2col via CMD, instead of hardcoded


tancheng:
https://github.com/tancheng/VectorCGRA/pull/301 provides im2col unit, but the input size and base addr are hardcoded at modeling time. In other words, though it is parameterizable, the parameterization happens when we generate the RTL.

Instead, we should allow user to program the base addr and input size during runtime, which requires the im2col unit be able to accept `CMD` and configure them.

tancheng:
fixed by https://github.com/tancheng/VectorCGRA/pull/334

PR #334, opened 2026-08-04
[refactor] Runtime-configure im2col via packed LAUNCH packet fields

zsjiang99:
Addresses issue #333: previously H, W, kH, kW, stride, in_base and the output SRAM base were all baked in at construct time, so the same synthesized RTL could only run one geometry. Now the engine reads them from the CMD_IM2COL_LAUNCH packet on every launch, so a single RTL can be reprogrammed at runtime.

Design (0 new CMDs):
  - Reuse the existing CMD_IM2COL_LAUNCH packet. Its fields carry the full config; the launch pulse and the config transfer are one and the same packet.
  - Layout (32-bit data field + 7-bit data_addr):
      data_addr          = dest_base
      data[ 5: 0]        = W        (6 bit)
      data[11: 6]        = H        (6 bit)
      data[17:12]        = in_base  (6 bit)
      data[20:18]        = kW       (3 bit)
      data[23:21]        = kH       (3 bit)
      data[27:24]        = log2_stride  (4 bit; stride = 1 << this)
  - pack_im2col_launch() helper in the engine module hides the bit
    layout from CPU code.

Engine interface:
  - start: InPort(b1)  ->  recv_cmd_pkt: RecvIfcRTL(IntraCgraPktType)
  - rdy is high only in S_IDLE, so LAUNCH cannot re-trigger mid-run.
  - Precomputes cfg_Hout / cfg_Wout on the LAUNCH edge so the S_EMIT loop-advance compares against plain registers each cycle.

Tests:
  fu/others/test/Im2colEngineRTL_test.py  5/5 pass
    - four geometry variants + a nonzero-dest_base check
  cgra/test/Im2colCgraE2E_test.py         3/3 pass (incl. Verilog gen)

### DIFF
--- a/cgra/test/Im2colCgraE2E_test.py
+++ b/cgra/test/Im2colCgraE2E_test.py
@@ -47,6 +47,7 @@ from pymtl3.stdlib.test_utils import (run_sim,
                                       config_model_with_cmdline_opts)
 
 from ..IntegratedIm2ColWithCgraRTL import IntegratedIm2ColWithCgraRTL
+from ...fu.others.Im2colEngineRTL import pack_im2col_launch
 from ...fu.double.SeqMulAdderRTL import SeqMulAdderRTL
 from ...fu.flexible.FlexibleFuRTL import FlexibleFuRTL
 from ...fu.float.FpAddRTL import FpAddRTL
@@ -146,7 +147,7 @@ class TestHarness(Component):
                 topology, controller2addr_map,
                 idTo2d_map, complete_signal_sink_out,
                 multi_cgra_rows, multi_cgra_columns, src_query_pkt,
-                engine_image, engine_geom, engine_scratch_mem_size):
+                engine_image, engine_scratch_mem_size):
 
     DataAddrType = mk_bits(clog2(data_mem_size_global))
     s.num_tiles  = width * height
@@ -164,10 +165,6 @@ class TestHarness(Component):
         FlexibleFuRTL, FuList, topology,
         controller2addr_map, idTo2d_map,
         engine_scratch_mem_size,
-        engine_geom['in_base'],
-        engine_geom['H'],        engine_geom['W'],
-        engine_geom['kH'],       engine_geom['kW'],
-        engine_geom['stride'],
         engine_image)
 
     cmp_fn = lambda a, b: (a.payload.data == b.payload.data and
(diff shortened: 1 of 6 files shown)


cgra/test/Im2colCgraE2E_test.py
```diff
@@ -369,12 +366,21 @@ def _run(pe_weights, expected_outputs,
...
-  launch_pkt = IntraCgraPktType(
-      payload = CgraPayloadType(CMD_IM2COL_LAUNCH))
+  # DMA-style trigger: prepend a CMD_IM2COL_LAUNCH packet whose fields
+  # carry the runtime im2col config. The controller forwards this single
+  # packet to send_to_im2col_engine_pkt; the engine parses the packed
+  # fields on the IDLE->READ edge.
+  launch_data = pack_im2col_launch(
+      H       = engine_geom['H'],
+      W       = engine_geom['W'],
+      in_base = engine_geom['in_base'],
```
tancheng: `in_base` -> `input_dram_base_addr`

hmm, I am actually not sure what do you mean by `in_base`... I thought the inputs are sequentially fed in. It doesn't need an "address"...

cgra/test/Im2colCgraE2E_test.py
```diff
@@ -369,12 +366,21 @@ def _run(pe_weights, expected_outputs,
...
+      kH      = engine_geom['kH'],
+      kW      = engine_geom['kW'],
+      stride  = engine_geom['stride'])
+  launch_pkt = IntraCgraPktType(payload = CgraPayloadType(
+      CMD_IM2COL_LAUNCH,
+      data      = DataType(launch_data, 1),
+      data_addr = engine_geom['dest_base']))
```
tancheng: `dest_base` -> `dst_sram_base_addr`

fu/others/Im2colEngineRTL.py
```diff
@@ -2,78 +2,132 @@
...
+
+_W_SHIFT           = 0
+_H_SHIFT           = _W_BITS                                            # 6
+_IN_BASE_SHIFT     = _H_SHIFT + _H_BITS                                 # 12
+_KW_SHIFT          = _IN_BASE_SHIFT + _IN_BASE_BITS                     # 18
+_KH_SHIFT          = _KW_SHIFT + _KW_BITS                               # 21
+_LOG2_STRIDE_SHIFT = _KH_SHIFT + _KH_BITS                               # 24
+
+
+def pack_im2col_launch(H, W, in_base, kH, kW, stride):
```
tancheng: I don't get this func. It seems you embed everything into a single msg. What if the data width is not long enough to hold all above commands/fields?

I originally thought we should have one cmd for each of above config, e.g.,
```
CMD_IM2COL_W
CMD_IM2COL_H
CMD_IM2COL_KW
CMD_IM2COL_KH
...
```

cgra/test/Im2colCgraE2E_test.py
zsjiang99: Yes, it's not necessary to use "in_base",  we can let it be "0" and use row, column to find the pixel address in input buffer.

cgra/test/Im2colCgraE2E_test.py
zsjiang99: Ok, I will fix it. Thanks.

fu/others/Im2colEngineRTL.py
zsjiang99: I see, it may be possible that data width is not long enough. Thanks for your suggestion. I will rewrite the code as you suggested.

zsjiang99: Hi Cheng, I just fixed and checked it. Thanks for your comments. Please let me know if you have any questions.

cgra/test/Im2colCgraE2E_test.py
```diff
@@ -366,22 +395,10 @@ def _run(pe_weights, expected_outputs,
...
-      stride  = engine_geom['stride'])
-  launch_pkt = IntraCgraPktType(payload = CgraPayloadType(
-      CMD_IM2COL_LAUNCH,
-      data      = DataType(launch_data, 1),
-      data_addr = engine_geom['dest_base']))
-  src_ctrl_pkt = [launch_pkt] + src_ctrl_pkt
+  # Prepend the im2col CONFIG + LAUNCH sequence. The controller forks
+  # each of these packets to send_to_im2col_engine_pkt; the engine
+  # latches config values in S_IDLE and starts on LAUNCH.
+  src_ctrl_pkt = _make_im2col_prologue(engine_geom) + src_ctrl_pkt
```
tancheng: `_make_im2col_prologue()` -> `_make_im2col_prepare_cmds()`

tancheng (approved):

cgra/test/Im2colCgraE2E_test.py
zsjiang99: Ok, got it. I have fixed it. Thanks.

Issue #59, opened 2025-01-05
[P3] Support flit for NoC pkg


tancheng:
There are two NoCs now, one for inter-CGRAs, and one for intra-CGRA (i.e., inter-tiles for delivering ctrl/const).

Instead of extending the pkg format for any novel signal, we can make one pkg be multi-flit, and fix the bandwidth. This would increase the complexity of the controller (requiring additional serializer/deserializer or assembler/disassembler) but very necessary.

yyan7223:
I need to figure out what does `flit` mean first.

tancheng:
Just some terminology in NoC.

flit vs. packet in NoC


ChatGPT said:
In a Network-on-Chip (NoC), flit and packet are fundamental units of data transmission, and understanding the distinction between them is crucial for grasping how data flows through a NoC.

🔹 Packet
A packet is the complete unit of data sent from a source to a destination.

It typically contains:

A header (with routing and control info),

A payload (the actual data),

And sometimes a tail (to indicate the end of the packet).

Think of it like a full letter (envelope + message).

🔹 Flit (Flow Control Unit)
A flit is a subdivision of a packet—the smallest unit that is handled by the flow control logic in the router.

Packets are broken down into flits for transmission, especially when a packet is too large to be sent in one cycle.

Common types of flits:

Header flit – carries routing information.

Body flits – carry the data (payload).

Tail flit – marks the end of the packet.

Example: A 128-bit packet might be broken into four 32-bit flits for transmission over a 32-bit-wide NoC link.

🔁 Relationship
Packet = multiple flits

Routers route at the flit level, but maintain packet-level semantics (e.g., they ensure ordering of flits within a packet).

⚙️ Why Flits?
Flow control is more efficient and fine-grained.

Enables pipelining of data in the NoC.

Helps deal with congestion and buffering more effectively.

tancheng:
@yyan7223 assigned to you in April :-)

yyan7223:
The design of serializer and its connection to `CgraRTL` and `MeshNetworkRTL` is shown below.

<img width="695" height="515" alt="Image" src="https://github.com/user-attachments/assets/3cab7f8c-c6c6-4408-b69c-aaa3a81b1b31" />

yyan7223:
The design of de-serializer and its connection to `CgraRTL` and `MeshNetworkRTL` is shown below.

[...]

PR #335, opened 2026-08-24
Serializer&Deserializer implementation and test.

yyan7223:
This PR is to support flit for NoC pkg (#59).
### Serializer design
c428bcfe-36ab-4732-b87a-586ce5d2a3eb

Serializer for converting a full NoCPktType packet into multiple NoCFlitType flits to be sent over the NoC. The packet payload is serialized from LSB to MSB. If the packet payload width is not an integer multiple of the flit payload width, the final flit is padded with zeros in its upper bits.

### Deserializer design
b5bcfa72-c440-4639-9e04-4c9c25c0b541
Deserializer for converting multiple NoCFlitType flits received from the NoC into a full NoCPktType packet. The packet payload is deserialized from LSB to MSB. If the packet payload width is not an integer multiple of the flit payload width, the padded upper bits of the final flit are discarded.

### DIFF
--- a/noc/test/DeserializerRTL_test.py
+++ b/noc/test/DeserializerRTL_test.py
@@ -1,0 +1,165 @@
+'''
+=========================================================================
+DeserializerRTL_test.py
+=========================================================================
+Simple test for DeserializerRTL.
+
+Author : Yufei Yang
+  Date : Aug 27, 2026
+'''
+
+from pymtl3 import *
+from pymtl3.stdlib.test_utils import config_model_with_cmdline_opts
+
+from ..DeserializerRTL import DeserializerRTL
+from ...lib.basic.val_rdy.SinkRTL import SinkRTL as TestSinkRTL
+from ...lib.basic.val_rdy.SourceRTL import SourceRTL as TestSrcRTL
+from ...lib.messages import *
+from ...lib.util.common import *
+
+
+#-------------------------------------------------------------------------
+# TestHarness
+#-------------------------------------------------------------------------
+
+class TestHarness(Component):
+
+  def construct(s, NoCPktType, NoCFlitType, flit_msgs, expected_pkt_msgs):
+
+    # Instantiates the DUT and the test source/sink.
+    s.dut = DeserializerRTL( NoCPktType, NoCFlitType )
+    s.src = TestSrcRTL( NoCFlitType, flit_msgs )
+
+    # Delays consuming the output packet to test backpressure on incoming flits.
+    s.sink = TestSinkRTL( NoCPktType, expected_pkt_msgs, initial_delay=10, interval_delay=10 )
+
+    # Connects the modules.
+    s.src.send //= s.dut.in_flit
+    s.dut.out_pkt //= s.sink.recv
+
+  def done(s):
+    return s.src.done() and s.sink.done()
+
+  def line_trace(s):
+    return s.dut.line_trace()
+
+
+#-------------------------------------------------------------------------
+# run_sim
+#-------------------------------------------------------------------------
+
+def run_sim(test_harness, max_cycles = 100):
+
+  # Creates a simulator.
+  test_harness.elaborate()
+  test_harness.apply( DefaultPassGroup() )
+  test_harness.sim_reset()
+
+  # Tracks whether the next flit is blocked while the current packet waits.
+  saw_blocked_next_flit = False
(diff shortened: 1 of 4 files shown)


tancheng:
add brief description and diagram in this PR's description section?

noc/test/DeserializerRTL_test.py
```diff
@@ -0,0 +1,165 @@
...
+
+  # Defines FlitType with a 32-bit width.
+  FlitType = mk_bits(flit_width)
+  NoCFlitType = mk_inter_cgra_pkt(
+    num_cgra_columns,
+    num_cgra_rows,
+    num_tiles,
+    num_rd_tiles,
+    FlitType
+  )
```
tancheng, line 122: Hmm, I think the packet->flit concept is different from my understand (based on NoC textbook).

I looked into https://github.com/tancheng/PyOCN/blob/d8ffacc64fb8dd24ce26e944d501016880302bbe/pymtl3_net/ocnlib/packets/MflitPacket.py, but found it is explicitly mentioned (non-translatable).

We either need to make it translatable, or hold on this PR. wdyt?

yyan7223: > Hmm, I think the packet->flit concept is different from my understand (based on NoC textbook).

I remember that you said flit head-body-tail before, do you mean this? Only head flit should contain the destination information. This may require the router to lock the routing direction after analysing the head flit, let the body flits flow along this direction, and unlock it after receiving the last flit, am I right?

> We either need to make it translatable, or hold on this PR. wdyt?

Don't worry, I'm happy to follow the standard flit concept to re-design the Serialized&Deserializer. Can you share the NoC textbook with me?

yyan7223: Actually I also think adding destination information for each flit is pretty redundant, but it requires no modification to other modules.

tancheng: https://dl.acm.org/doi/book/10.5555/3134175

The flit concept should be invisible to any IP (e.g., CGRA), it should be handled by the NoC itself. (maybe only serializer or deserializer needs a bit refactoring, but hopefully not)

yyan7223: Hi @tancheng ,

Thank you for recommending this book. After reading it with some help from GPT, I have three questions that I’d like to discuss with you.

(1) According to the book, a flit can be represented with three main fields: `Type`, `VCID`, and `Data`. The PyOCN design https://github.com/tancheng/PyOCN/blob/d8ffacc64fb8dd24ce26e944d501016880302bbe/pymtl3_net/ocnlib/packets/MflitPacket.py you mentioned also supports phits and includes additional fields such as `flit_idx` and `nflits`.

However, the book states at the end of page 59:
_“To date, in on-chip networks, flits are composed of a single phit and are the smallest subdivision of a message due to wide on-chip channels.”_

In our case, would it therefore be reasonable to simplify `NoCFlitType` to contain only `Type`, `VCID`, and `Data`?

(2) My understanding from the book is that the VC used by a packet is allocated hop by hop. For example, before a flit is sent from Router 0 to Router 1, Router 0 selects an available input VC at Router 1, and the outgoing flit carries the corresponding VCID. The body and tail flits then inherit the VC selected for the head flit.

After looking through the PyOCN implementation with GPT, I could not find an explicit VC allocator that dynamically selects and updates the VCID in this way. Do we need to add this mechanism in our implementation? If so, we may also need an injection VC allocator/selector before Router 0 to assign the initial VCID to the flits generated by the Serializer (Or maybe put the VC allocator just in Serializer).

(3) Should the Deserializer also be VC-aware? Since flits from different packets may be interleaved on the same physical link when multiple VCs are used, a single Deserializer state may not be sufficient to reconstruct multiple packets correctly. Would it make sense for the Deserializer to maintain separate state for different VCs, or should the ejection side first select one VC and drain the whole packet before sending it to the Deserializer?

WDYT?

tancheng: When Bill Dally first developed these concepts in the 1980s and 1990s, he was designing interconnection networks for massive supercomputers (like the Cray T3D and T3E) where physical wires were tightly constrained, meaning flits > phits.

But things changed today.

All ur comments make sense. Feel free to propose your own design. You can fork PyOCN (the version we used in VectorCGRA).

PR #338, opened 2026-09-08
Fix predicated-off stores in MemUnitRTL

HobbitQia:
## Problem

`OPT_STR` asserts both memory write valid signals whenever its input tokens are available, even when the address or data predicate is zero. Although the output data carries the combined predicate, the downstream memory writes on the valid handshake without checking that predicate. A predicated-off store can therefore overwrite a valid result.

This was investigated while checking incorrect first outputs in GEMV and AXPY reported in [coredac/CGRA-SoC#3](https://github.com/coredac/CGRA-SoC/issues/3).

## Reproduction

The new `test_store_predicate` regression uses the existing MemUnit and data-memory harness:

1. Store `40` at address `0` with both predicates set.
2. Attempt to store `99` at the same address with either the address predicate or the data predicate cleared.
3. Load address `0` and expect `40` with predicate `1`.

With the regression present, run from the repository root:

```bash
pytest fu/single/test/MemUnitRTL_test.py -k store_predicate -q --tb=short
```

Both variants fail against the unmodified `master` implementation at `ac96a7a`: the load returns `99` with predicate `0` instead of preserving `40` with predicate `1`.

## Fix

Gate both `to_mem_waddr.val` and `to_mem_wdata.val` with the address and data predicates, matching the existing `OPT_STR_CONST` behavior. Keep operand and control readiness unchanged so predicated-off stores can consume their inputs and complete without issuing a memory write.

## Tests

- All four MemUnit tests pass after the fix, including both regression variants.
- Both new regression variants pass with `--test-verilog`. Local validation used an explicit Verilator C++ prefix to resolve the installed PyMTL/Verilator naming mismatch.
- Applying the same MemUnit patch to an isolated `kernel-submit` checkout at `57f0bda` restores complete output correctness for GEMV (`30, 70, 110, 150`) and AXPY (`40` through `100`, step `4`). These kernel checks inspect every output, including the first.

The kernel results above are from `kernel-submit`, not latest `master`. The borrowed GEMV fixture does not complete within 600 cycles on latest `master`, both before and after this fix. Full-suite CI and Chipyard validation have not been run.

### DIFF
--- a/fu/single/MemUnitRTL.py
+++ b/fu/single/MemUnitRTL.py
@@ -199,12 +199,16 @@ class MemUnitRTL(Component):
           s.recv_in[s.in0_idx].rdy @= s.recv_all_val & s.to_mem_waddr.rdy & s.to_mem_wdata.rdy
           s.recv_in[s.in1_idx].rdy @= s.recv_all_val & s.to_mem_waddr.rdy & s.to_mem_wdata.rdy
           s.to_mem_waddr.msg @= AddrType(s.recv_in[s.in0_idx].msg.payload[0:AddrType.nbits])
-          s.to_mem_waddr.val @= s.recv_all_val
+          s.to_mem_waddr.val @= s.recv_all_val & \
+                                s.recv_in[s.in0_idx].msg.predicate & \
+                                s.recv_in[s.in1_idx].msg.predicate
           s.to_mem_wdata.msg @= s.recv_in[s.in1_idx].msg
           s.to_mem_wdata.msg.predicate @= s.recv_in[s.in0_idx].msg.predicate & \
                                           s.recv_in[s.in1_idx].msg.predicate & \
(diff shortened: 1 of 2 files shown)


fu/single/MemUnitRTL.py
tancheng, line 219: does `OPT_STR_CONST` also needs fix? i saw it already has:

```
          s.to_mem_wdata.val @= s.recv_all_val & \
                                s.recv_in[s.in0_idx].msg.predicate & \
                                s.recv_const.msg.predicate
```

tancheng (approved):

fu/single/MemUnitRTL.py
HobbitQia: This `OPT_STR_CONST` is already correct and I think no fix needs to be done~

Issue #339, opened 2026-09-15
Include from-YAML kernel tests in CI


HobbitQia:
Currently, pytest.ini collects only *_test.py, excluding files such as `CgraRTL_gemv_test_from_yaml.py` and `CgraRTL_histogram_test_from_yaml.py`. The workflow does not explicitly run these tests.

  Local PyMTL runs after merging upstream c9b200ff showed:

  - GEMV stalls without completing.
  - Histogram produces [4,5,5,5] instead of [5,5,5,5].

Maybe we can include these tests to CI so that we can check their correctness?

tancheng:
Thanks @HobbitQia, sure, we definitely wanna include them. But how to fix?

@Jackcuii these tests are failed and are excluded in your PR?

HobbitQia:
> Thanks [@HobbitQia](https://github.com/HobbitQia), sure, we definitely wanna include them. But how to fix?
> 
> [@Jackcuii](https://github.com/Jackcuii) these tests are failed and are excluded in your PR?

My previous PR #338 should partially solve this problem, including GEMV/AXPY. However, it seems that PR #336 introduced another bug of token consumption mechanism in registers, so GEMV test failed with deadlock...(I tried previous version, which can pass).

Regarding Histogram, the error should come from `ExclusiveDivRTL` which produces the wrong results.

tancheng:
Thanks @HobbitQia, can you try to fix each test with a new PR? For the GEMV, would the bug be fixed by reverting the code changes related to registers?

And plz fix `ExclusiveDivRTL` with another PR.

HobbitQia:
> Thanks [@HobbitQia](https://github.com/HobbitQia), can you try to fix each test with a new PR? For the GEMV, would the bug be fixed by reverting the code changes related to registers?
> 
> And plz fix `ExclusiveDivRTL` with another PR.

I can try them these days.

HobbitQia:
@tancheng The GEMV failure appears to come from the register token consumption semantics introduced in #322, which remain in place after reverting #336. A register’s token is invalidated when a control step that reads it completes, even if a later step still needs the same value.

For example, core 5 reads register `$1` at step 10, then needs that value again at step 0 of the next iteration. Completing step 10 clears its `token_valid` bit, so the subsequent read cannot provide a valid operand, even though the register still contains the required value. This causes GEMV to stall without completion.

[...]

PR #340, opened 2026-09-17
Fix register value retention and run YAML kernels in CI

HobbitQia:
## Summary

- Add a per-bank `read_reg_retain` control bit and generator analysis to preserve register values needed by later control steps, fixing the GEMV stall.
- Add an explicit GitHub Actions step for YAML kernel tests and use `pathlib` to resolve YAML paths independently of the working directory.
- Skip SpMV since there are unsupported operations (e.g., `ICMP_ULT`, constant `AND`) in YAML.

## Design

The per-bank `retain` control bit: when set, completing a read preserves the token. The YAML generator would set this bit when the same register must be read again before being overwritten. The final read would consume the token normally.

Regarding the retain bit, for example, GEMV reads $1 at step 10 and needs the same value again at step 0 of the next iteration. At the end of step 10, the bank only sees the current control and handshake signals. Those signals do not tell it whether this was the final use, so they are insufficient to infer retain automatically.

## Validation

YAML tests run from `build/`: **7 passed, 1 skipped**. Final workflow collection and all YAML paths were also verified.

It's worth noting that for now GEMM currently checks completion only, and its output of GEMM results is incorrect.

Ref #339


tancheng:
@HobbitQia thanks for the fix~!
 - Can you plz provide an example showing why we need this retain? (in the PR description section)
 - Can you plz explain when (depending on what) the retain ctrl bit would be set? It won't cause trouble for the invalid signal introduced by previous PR?
 - The script generator update in this repo is enough? We don't need to update the compiler to provide more info?
 - The previously introduced yaml was not include in our test, but now they are included by this PR, and only one is skipped, right?
 - You have plan to fix the GEMM in another PR?

HobbitQia:
> * Can you plz provide an example showing why we need this retain? (in the PR description section)

I added the illustration in Issue #339 into this PR's description.

> * Can you plz explain when (depending on what) the retain ctrl bit would be set? It won't cause trouble for the invalid signal introduced by previous PR?

The generator sets retain ctrl bit when the next access to the same register in the same bank reads it before overwriting it, including across the iteration boundary. For now it passes all tests including previous tests. I think as long as the YAML generator can correctly recognize the last use of the value, the behavior of retain is correct. 

> * The script generator update in this repo is enough? We don't need to update the compiler to provide more info?

For now I think it's enough.


> * The previously introduced yaml was not include in our test, but now they are included by this PR, and only one is skipped, right?

Yes.

> * You have plan to fix the GEMM in another PR?

It seems that GEMM's results are incorrect in recent versions of repo. So I may need more time to locate the problem...

tancheng:
> I added the illustration in Issue https://github.com/tancheng/VectorCGRA/issues/339 into this PR's description.

Where is the "illustration"? you mean the "Design" section?

> The generator sets retain ctrl bit when the next access to the same register in the same bank reads it before overwriting it, including across the iteration boundary.

The token introduced by #322 is to handle the case of (`$north_port, $reg0 = inc $reg0`, sometimes, `$north_port` won't be ready, so we need token on `$reg0`'s bank). `retain` won't affect it?

HobbitQia:
Sorry for my late reply...

> Where is the "illustration"? you mean the "Design" section?

Yes. This part is copied from the issue and serves as the illustration. If you think anything else should be added, I am glad to do it.

> The token introduced by #322 is to handle the case of (`$north_port, $reg0 = inc $reg0`, sometimes, `$north_port` won't be ready, so we need token on `$reg0`'s bank). `retain` won't affect it?

Current fix also passed the tests in PR #322 and Issue #321. I think there will be no influence since it will only control when the token should be kept after read, rather than when it should be written back.

tancheng:
> Current fix also passed the tests in PR https://github.com/tancheng/VectorCGRA/pull/322 and Issue https://github.com/tancheng/VectorCGRA/issues/321. I think there will be no influence since it will only control when the token should be kept after read, rather than when it should be written back.

Will the PR works when the two cases happen together? i.e., `$north_port, $reg0 = inc $reg0` followed by `$south_port = add $reg0, $west_port`.

